PR #1, opened 2021-11-15
feat: setup the repo with CI using github actions

wendychiang:
Build can be found here: https://github.com/godaddy/sample-size/runs/4218565006?check_suite_focus=true


.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+jobs:
+  tests:
+    name: Checks of unit tests and format checks
+    runs-on: "${{ matrix.os }}"
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: [3.7, 3.8, 3.9]
```
wendychiang, line 17: although engineer will only test it locally with a single version (now suggest as 3.7.8), guess still better to test it with any python version > 3.7?

CONTRIBUTING.md
```diff
@@ -107,10 +107,59 @@ approved, other developers with commit access may commit it for the author. When
 doing so, it is important to retain correct attribution of the contribution.
 Generally speaking, Git handles attribution automatically.
 
+## Development
+
+### Local setup
+
+Install [pyenv](https://github.com/pyenv/pyenv#installation) and [poetry](https://python-poetry.org/docs/#installation)
+
+Install and set python 3.7.8 with pyenv:
```
wendychiang, line 116: can also use a newer version though

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
+name: Run checks on pull requests and pushes on main branch
+
+on:
+  pull_request: {}
+  push:
+    branches: [main]
```
wendychiang, line 6: currently any pull request and any PR merges to `main` will trigger this workflow. If we will need new steps when merging to main e.g publish version. we can add new yml file and separate them in the future

sample_size/sample_size.py
```diff
@@ -0,0 +1,17 @@
+from typing import List
```
wendychiang, line 1: simply test NumPy and statsmodels are imported correctly, this will be directly overridden soon

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+
+jobs:
+  tests:
+    name: Checks of unit tests and format checks
+    runs-on: "${{ matrix.os }}"
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      fail-fast: false
```
vjohnson1-godaddy, line 14: Would setting [this to true](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions) be closer to our usual paradigm?

CONTRIBUTING.md
```diff
@@ -107,10 +107,59 @@ approved, other developers with commit access may commit it for the author. When
...
+
+```console
+pyenv install 3.7.8
+pyenv local 3.7.8
+```
+
```
vjohnson1-godaddy, line 122:
````suggestion

To validate that your poetry venv is using the correct python version
```console
poetry env info
```

````

File: pyproject.toml [hunk 7]
```diff
@@ -0,0 +1,88 @@
...
+mypy = "^0.910"
+pytest = "^6.2.2"
+pytest-sugar = "^0.9.4"
+pytest-cov = "^2.11.1"
+
+[tool.poetry.scripts]
+qa = "poetry_scripts:qa"
+test = "poetry_scripts:test"
+format = "poetry_scripts:format_fix"
+format_check = "poetry_scripts:format_check"
```
vjohnson1-godaddy, line 25: nit: I feel like CLI usually use hyphen instead of underscores for some reason?
```suggestion
format-check = "poetry_scripts:format_check"
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
+To help us get this pull request reviewed and merged quickly, please be sure to include the following items:
+
+* [ ] Tests (if applicable)
+* [ ] Documentation (if applicable)
+* [ ] A full explanation here in the PR description of the work done
```
vjohnson1-godaddy, line 5: Just opining here. Feel free to skip

```suggestion
* [ ] A full explanation here in the PR description of the work done and statistical methods used
```

pyproject.toml
wendychiang: sure, although that is actually follow godaddy example :D

.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
...
+
+* [ ] Bugfix
+* [ ] Feature
+* [ ] Code style update (formatting, local variables)
+* [ ] Refactoring (no functional changes, no api changes)
+* [ ] Build related changes
+* [ ] CI related changes
+* [ ] Documentation content changes
+* [ ] Tests
+* [ ] Other
```
dpandey-godaddy, line 18: Do we need `Other` here? What are we missing?

.github/workflows/ci.yml
wendychiang: yeah the behavior is to cancel other parallel jobs, in this case it would be different python versions, and windows mac system jobs. (guess it's a bit different than our existing jenkins job which is just one time run with a single environment)
I feel this would be helpful to be false to directly know which jobs are falling when there are multiple ones failed. For example, if there are 3 jobs failing out of total 9 jobs. we can know it directly instead of seeing 1 failed, try to fix that one. And potentially on the next try, the other job failed first so we couldn't know if the one we are trying to fix works or not.
But yeah if we still think fail first is what we want, i can change

vjohnson1-godaddy: Your comment makes sense and I now agree with `fail-fast: false`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+        shell: bash
+        run: timeout 10s poetry run pip --version || rm -rf .venv
+
+      - name: Install dependencies
+        shell: bash
+        run: poetry install
+
+      - name: Format check
+        shell: bash
+        run: poetry run format_check
```
dpandey-godaddy, line 80: These needs to be updated as well

.github/pull_request_template.md
wendychiang: maybe we didn't miss anything, but doesn't hurt to have it just in case? e.g adding code owners or something

.github/workflows/ci.yml
tarkatronic: I would even recommend adding `3.10` to this list, as it was released on October 4!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
+name: Run checks on pull requests and pushes on main branch
+
+on:
+  pull_request: {}
+  push:
+    branches: [main]
+
+jobs:
+  tests:
+    name: Checks of unit tests and format checks
```
tarkatronic, line 10: You might want to consider a shorter name here, for readability. If you look here, you can see that the names get truncated, and it's difficult to tell which run is which platform/Python version: https://github.com/godaddy/sample-size/runs/4230160107?check_suite_focus=true

One thing you might consider is something similar to what I did with `tartufo`, splitting the unit tests and linters into distinct jobs. That also allows them to fail separately, so that the contributor knows more easily whether a test or a linter failed. 

See here: https://github.com/godaddy/tartufo/blob/3b4acef809158676ebf7dfa8b93ae6f0a525b199/.github/workflows/ci.yml#L11
And here: https://github.com/godaddy/tartufo/blob/3b4acef809158676ebf7dfa8b93ae6f0a525b199/.github/workflows/ci.yml#L62

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Configure poetry
+        shell: bash
+        run: poetry config virtualenvs.in-project true
+
+      - name: Set up cache
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: .venv
```
tarkatronic, line 66: You should make sure to also cache `${{ matrix.path }}`, since you've gone to the work of configuring it up above.

CONTRIBUTING.md
tarkatronic: As long as the contributor is using one of the "supported" Python versions (one of the ones tested in your CI), does it matter which specific one they use? What if they want to use 3.7.7? Or 3.7.12? For that matter, what if they don't want to use pyenv? It should be totally legitimate to use packages straight from python.org, or from Homebrew, or from the `asdf` package manager for example.

The goal here should be to reduce friction for potential contributors, and I could see dictating specific tools and versions like this as increasing friction, making it more difficult for a new contributor to get started, meaning they may decide it's not worth it.

CONTRIBUTING.md
```diff
@@ -107,10 +107,65 @@ approved, other developers with commit access may commit it for the author. When
...
+To validate that your poetry venv is using the correct python version
```
tarkatronic, line 123: Having said all that above, it's probably worth linking to the Poetry installation doc in case a user isn't familiar with Poetry.

If you'd like an example of a very minimal developer setup guide, this is what we've written up for `tartufo`: https://tartufo.readthedocs.io/en/stable/CONTRIBUTING.html#writing-code

wendychiang: poetry doc is linked in the beginning of this section, but yeah i agree with the comment. This current setup is the way our team does, i will keep it but add a note saying pyenv is just a recommended way but not a requirement to work on it

CONTRIBUTING.md
wendychiang: updated the doc

.github/workflows/ci.yml
wendychiang: added

vjohnson1-godaddy (approved):

CONTRIBUTING.md
```diff
@@ -107,10 +107,85 @@ approved, other developers with commit access may commit it for the author. When
...
+Install pyenv [here]https://github.com/pyenv/pyenv#installation)
+
+Install and set python version with pyenv:
+
+```console
+pyenv install 3.7.8
+pyenv local 3.7.8
+```
+
+Note: You are welcome to use any python version >= 3.7.1
```
chuxiaofdn, line 139: do we want to be more specific on the versions we support?  3.7 to 3.10?

wendychiang: sure, just updated, it needs to actually >= 3.7.1 for pandas to work for 3.10

.github/pull_request_template.md
dpandey-godaddy: we may have to revisit, if there are many prs specified as **Other**

chuxiaofdn (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Format check
+        shell: bash
+        run: poetry run format-check
+
+      - name: Lint
+        shell: bash
+        run: poetry run lint
+
+      - name: Lint
```
satish-ravi, line 83: Should this be Type check?

CONTRIBUTING.md
```diff
@@ -107,10 +107,85 @@ approved, other developers with commit access may commit it for the author. When
...
+### Setup with pyenv
+
+It is recommended to use pyenv to switch between Python versions:
+
+Install pyenv [here]https://github.com/pyenv/pyenv#installation)
+
+Install and set python version with pyenv:
+
+```console
+pyenv install 3.7.8
```
satish-ravi, line 135: Maybe use a more recent version?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Get full Python version
+        id: full-python-version
+        shell: bash
+        run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
+
+      - name: Bootstrap poetry
+        shell: bash
+        run: |
+          curl -sL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py \
+            | python - -y ${{ matrix.bootstrap-args }}
```
satish-ravi, line 44: Is it worth considering something like https://github.com/abatilo/actions-poetry?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+  tests:
+    name: Format and Tests
+    runs-on: "${{ matrix.os }}"
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        experimental: [false]
```
satish-ravi, line 18: what does `experimental` mean here?

wendychiang: if setting it to true, it will run without invalidating the whole workflow if these jobs happen to fail due to an unstable dependency. If that is confusing, I'll remove since it is false here

PR #2, opened 2021-11-17
Initial check in for sample size calculator

chuxiaofdn:

* Initial check in for sample size calculators
* Initial check in for utils scripts to take user input
* Unit tests included


sample_size/utils/sample_size_run.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Calculate sample size based on user inputs for
+        1. metric type: Boolean, Numeric, or Ratio (case insensitive)
+        2. power analysis parameters: alpha, power, and mde
+        3. variable input:
+            * Boolean: probability
+            * Numeric: mean and variance
+            * Ratio: mean and variance of numerator and denominator and their covariance
+
+    For future use in Hivemind, this is a sample code to calculate sample sizes with given parameters
```
wendychiang, line 16: probably better to not mention any internal thing since this is in a public place. However, the below implementation can probably be added to README.md for the simple usage overview. something like https://github.com/PyGithub/PyGithub#simple-demo

sample_size/utils/utils.py
```diff
@@ -0,0 +1,94 @@
...
+def is_float(value: str) -> float:
+    try:
+        float(value)
+        return True
+    except ValueError:
+        return False
+
+
+def get_raw_input(text: str) -> str:
+    return input(text)
```
wendychiang, line 18: if this function is just returning input(), we don't actually need an extra function? can just do text() directly when we need it?

sample_size/utils/sample_size_run.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Get calculator
+    calculator = SampleSizeCalculators(
+        power_analysis_parameters.mde, power_analysis_parameters.alpha, power_analysis_parameters.power
+    )
+
+    # Get and print sample size based on variable and power analysis parameters
+    sample_size = get_sample_size(variable_name, calculator)
+    print(sample_size)
+    print("\n Sample size needed in each group: {:.3f}".format(sample_size))
```
wendychiang, line 43: how about putting these implementation under a `main()` function then we are able to add tests on this main function.

sample_size/utils/sample_size_run.py
```diff
@@ -0,0 +1,43 @@
+from sample_size.sample_size_calculator.sample_size_calculator import SampleSizeCalculators
```
wendychiang, line 1: since this implementation is independent of the sample size calculator, how about putting it outside, putting it under utils directory seems like it has some functions that is needed by `sample_size_calculator`
we can add a scripts directory, something like
```
sample-size-repo/scripts/run/sample_size_run.py
                             utils.py
                             __init__.py
```

wendychiang (commented):
probably want to avoid committing `.Ds_Store` file
haven't reviewed all detailed implementation, add some comments first

sample_size/utils/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        raise Exception("Error: Unexpected variable name. Please use Boolean, Numeric, or Ratio.")
+
+
+def get_power_analysis_input() -> PowerAnalysisParameters:
+
+    parameters = PowerAnalysisParameters()
+    alpha = get_alpha(get_input("alpha (default 0.05)", allow_na=True))
+    if alpha:
+        parameters.alpha = alpha
+    parameters.mde = get_input("minimum detectable effect")
```
wendychiang, line 77: might be better to mention this is a relative number or absolute number

sample_size/sample_size_calculator/sample_size_calculator.py
```diff
@@ -0,0 +1,105 @@
...
+                power=self.power_analysis_parameters.power,
+                ratio=1,
+                alternative="two-sided",
+            )
+        )
+
+        return sample_size
+
+
+class SampleSizeCalculators:
```
wendychiang, line 70: how about making it in a way that doesn't require a lot of refactoring once we support multiple testing (support multiple metrics)? for both the package and the usage. some suggestions below. With that we don't need to refactor much later on.
## Package
```python
class SampleSizeCalculator(object):

    def __init__(self, alpha, power, comparison=2):
        self.alpha = alpha
        self.power = power
        self.comparison = comparison # just for future use, don't need to be included now
        self.boolean_metrics: List[Boolean] = []
        self.numeric_metrics: List[Numeric] = []
        self.ratio_metrics: List[Ratio] = []

    
    def register_bool_metric(self, mde, probability):
        variable = Boolean(mde, probability)
        self.boolean_metrics.append(variable)
    
    def register_ratio_metric(self, mde, numerator_mean, numerator_variance, denominator_mean, denominator_variance, covariance):
        variable = Ratio(mde, numerator_mean, numerator_variance, denominator_mean, denominator_variance, covariance)
        self.ratio_metrics.append(variable)
    
    def register_numeric_metric(self, mde, mean, variance):
        variable = Numeric(mde, mean, variance)
        self.numeric_metrics.append(variable)
    
    def get_overall_sample_size(self):
        # temporarily directly return after calculating the non-null one since it only support single metric now
        if self.boolean_metrics:
            for metric in self.boolean_metrics:
                return self.__get_single_sample_size(metric.variance, metric.mde, NormalIndPower)
        if self.numeric_metrics:
            for metric in self.numeric_metrics:
                return self.__get_single_sample_size(metric.variance, metric.mde, TTestIndPower)
        if self.ratio_metrics:
            for metric in self.ratio_metrics:
                return self.__get_single_sample_size(metric.variance, metric.mde, TTestIndPower)
        
        # In the future, this function will use self.comparison to apply some corrections
        # and eventually return an overall sample size
        # e.g return self.correct_sample_size(sizes, self.comparisoin)

    def __get_single_sample_size(self, mde, variance, power_analysis_type):
        effect_size = mde / float(np.sqrt(variance))
        sample_size = power_analysis_type().solve_power(effect_size, self.alpha, self.power, ratio=1, alternative="two-sided",)
        return sample_size
```

## Usage
```python
calculator = SampleSizeCalculator(alpha, power, comparison)
for metric in metrics:
    if isinstance(metric, Boolean):
        calculator.register_bool_metric(metric.mde, metric.value)
    elif isinstance(metric, Numeric):
        calculator.register_numeric_metric(metric.mde, metric.value, metric.variance)
    elif isinstance(metric, Ratio):
        calculator.register_ratio_metric(...)
sample_size = calculator.get_overall_sample_size()
# Note: the Boolean, Numeric, Ratio here 
# are different than the ones setup in this python package
```

sample_size/sample_size_calculator/sample_size_calculator.py
```diff
@@ -0,0 +1,105 @@
...
+        sample_size_calculator = BaseSampleSizeCalculator(variance, power_analysis_type, self.power_analysis_parameters)
+        return sample_size_calculator.get_base_sample_size()
+
+    def get_boolean_sample_size(self, probability: float) -> float:
+        bool_obj = Boolean(probability)
+        return self.get_sample_size(bool_obj.variance, NormalIndPower)
+
+    def get_numeric_sample_size(self, mean: float, variance: float) -> float:
+        numeric_obj = Numeric(mean, variance)
+        return self.get_sample_size(numeric_obj.variance, TTestIndPower)
```
wendychiang, line 94: my own question: so that means for numeric, we don't even need the `mean` right?

wendychiang (commented):
main suggestion on the calculator class design. also maybe can have separate PRs - one for the calculator, one for the script to ease the review process

sample_size/utils/utils.py
chuxiaofdn: I didn't find an easy way to test input() or mock input(). `get_raw_input` is created to avoid calling input() directly. Any suggestions on the test?

sample_size/sample_size_calculator/sample_size_calculator.py
chuxiaofdn: yeah you are right. I'll remove mean from numeric.

PR #3, opened 2021-11-20
ci: add release workflow and instructions

wendychiang:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
...
+    - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+    - name: Set up Python 3.7
+      uses: actions/setup-python@3105fb18c05ddd93efea5f9e0bef7a03a6e9e7df
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run:  |
+        pip install --upgrade pip
+        pip install --upgrade poetry
+    - name: Publish to Test-PyPI with Poetry
```
wendychiang, line 23: we don't really need to publish to test-pypi, so we can remove this if we want.
I've tested, both steps should work when a Release is published

.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
+name: publish
```
wendychiang, line 1: this file is almost the same as tartufo one (another godaddy opensource python package) [here](https://github.com/godaddy/tartufo/blob/main/.github/workflows/publish.yml), without publishing to docker hub.

CONTRIBUTING.md
```diff
@@ -187,6 +187,47 @@ poetry run qa  # run all checks
...
+Once your PR has been merged, the final piece is to actually create the new release.
+
+1. Go to the [sample-size releases page](https://github.com/godaddy/sample-size/releases) and click on `Create a new release`.
+
+1. Enter an appropriate tag version or create a new tag. Please prefix the version names with the letter `v`
+
+    ![release-tag](./docs/source/_static/release-tag.png)
+
+
+1. Title the release. Generally these would just be in the form `Sample-Size 0.1.0`
```
wendychiang, line 223: or we could do `v0.1.0`, I don't feel strongly. But include the package name is the same behavior that [boto3](https://github.com/boto/boto3/releases), [pandas](https://github.com/pandas-dev/pandas/releases) are doing

.github/workflows/publish.yml
satish-ravi: probably skip this publish?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+    - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+    - name: Set up Python 3.7
+      uses: actions/setup-python@3105fb18c05ddd93efea5f9e0bef7a03a6e9e7df
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run:  |
+        pip install --upgrade pip
+        pip install --upgrade poetry
```
satish-ravi, line 22: This poetry install seems inconsistent with the ci one. Should we make it consistent?

CONTRIBUTING.md
```diff
@@ -187,6 +187,47 @@ poetry run qa  # run all checks
 
 Ensure that your contribution follows the [PEP 8 – Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/) style guide with respect to patterns, naming, documentation and testing.
 
+## Issuing a New Release
```
satish-ravi, line 190: Should this be on a different file. Not sure if this makes sense in CONTRIBUTING.md

wendychiang: i'm not sure the standard way though, didn't find many examples. `tartufo` put them with contributing doc https://tartufo.readthedocs.io/en/latest/CONTRIBUTING.html, pytighub has a [DEPLOY.md](https://github.com/PyGithub/PyGithub/blob/master/DEPLOY.md). anyways, I moved them to `RELEASE.md` now

.github/workflows/publish.yml
wendychiang: update ci.yml to use pip now

satish-ravi (approved):

chuxiaofdn (approved):

PR #4, opened 2021-11-30
Add sample size calculator

chuxiaofdn:
The prior [PR](https://github.com/godaddy/sample-size/pull/2) has been split into two PRs for better review. 
This PR contains the calculator, variable classes, and the related tests.

* [ ] Tests (if applicable)

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [x] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)


sample_size/variables.py
```diff
@@ -0,0 +1,52 @@
+class Boolean:
```
wendychiang, line 1: follow the other comment, how about putting `mde` here as a variable then we can directly using these classes without PowerAnalysisParameters?

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,106 @@
...
+
+from sample_size.variables import Boolean
+from sample_size.variables import Numeric
+from sample_size.variables import Ratio
+
+DEFAULT_ALPHA = 0.05
+DEFAULT_POWER = 0.8
+
+
+class PowerAnalysisParameters:
```
vjohnson1-godaddy, line 16: I notice we use this class in the following way:

```python3
metric = PowerAnalysisParameters(variable.variance, mde, self.alpha, self.power)
```

I wonder if it might make more sense to call this class `Metric`? Jumping off of that, perhaps a metric would simply have a `Variable` (`Boolean`, `Ratio`, etc.) and an MDE? `alpha` and `power` seem to already be handled by the `SampleSizeCalculator` class?

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        return sample_size
+
+    def get_overall_sample_size(self) -> float:
+        # Supports the sample size calculation for single metric now.
+        # The current structure is set up to support multiple metrics in the future.
+        sample_size = float("nan")
+        if self.boolean_metrics:
+            for metric in self.boolean_metrics:
+                sample_size = self._get_single_sample_size(metric, NormalIndPower)
```
vjohnson1-godaddy, line 98: Will a particular metric ever have multiple `power_analysis_types`? Wonder if this could be represented in the `Boolean`, `Ratio`, and `Proportion` classes? If that was the case, this entire method could be represented as 

```python3
 def get_overall_sample_size(self) -> float:
        return max(map(self._get_single_sample_size, self.metrics))
```

because we would no longer need to differentiate `self.numeric_metrics`, `self.boolean_metrics`, and `self.ratio_metrics`

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,106 @@
...
+        self.alpha = alpha
+        self.power = power
+        self.boolean_metrics: List[PowerAnalysisParameters] = []
+        self.numeric_metrics: List[PowerAnalysisParameters] = []
+        self.ratio_metrics: List[PowerAnalysisParameters] = []
+
+    def register_bool_metric(self, mde: float, probability: float) -> None:
+        variable = Boolean(probability)
+        metric = PowerAnalysisParameters(variable.variance, mde, self.alpha, self.power)
+        self.boolean_metrics.append(metric)
```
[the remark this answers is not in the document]
chuxiaofdn: when we plan to support multiple metrics per calculator, the mde per metric will be different. Their alpha and power should be the same.  I can change PowerAnalysisParameters to house variable and MDE only, and call it a different name.

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @staticmethod
+    def _get_single_sample_size(metric: PowerAnalysisParameters, power_analysis_type: Any) -> float:
+        effect_size = metric.mde / float(np.sqrt(metric.metric_variance))
+        power_analysis = power_analysis_type()
+        sample_size = int(
+            power_analysis.solve_power(
+                effect_size=effect_size,
+                alpha=metric.alpha,
+                power=metric.power,
```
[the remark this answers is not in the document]
chuxiaofdn: related to the above comment. self.alpha and metric.alpha are currently the same. Can update this together with the PowerAnalysisParameters refactor.

sample_size/variables.py
chuxiaofdn: I hesitate to put `mde` in variable classes since it is a parameter for the calculator not variable.

sample_size/sample_size_calculator.py
chuxiaofdn: Yeah that's what I plan to do if no objections.

sample_size/sample_size_calculator.py
chuxiaofdn: Good question. For now, we support single metric per calculator, I don't think we will ever introduce multiple power analysis type per variable. We may need multiple correction methods when we support multiple metrics per calculator. I can put the default single metric power analysis type in variable class to make this look more obvious.

sample_size/variables.py
wendychiang: oh ok, i was interpreting each variable as a metric though (it doesn't seem that variable will be used for any other purpose outside of as a "metric" in this repo). and each Variable object can only have one MDE

chuxiaofdn: Good point. I can rename the variable classes and let them house mde.

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,80 @@
...
+                alternative="two-sided",
+            )
+        )
+        return sample_size
+
+    def get_overall_sample_size(self) -> float:
+        # Supports the sample size calculation for single metric now.
+        # The current structure is set up to support multiple metrics in the future.
+        sample_size = float("nan")
+        if self.boolean_metrics:
```
wendychiang, line 70: since `default_power_analysis_type` is under each variable class now, it looks like we can actually just have single variable `self.metrics`. but not sure if the current setting (three `self.{type}_metric`s) will be easier for multiple metrics correction later on.
Same idea as the comment that @vjohnson1-godaddy has https://github.com/godaddy/sample-size/pull/4#discussion_r759654663

chuxiaofdn: Yeah it depends on whether we need to apply multiple metric correction based on the metric type. How about we keep it like this and refactor (if necessary) after the research is completed?

wendychiang: sounds good, can you add a Note like ` consider having only one self.metrics` , thanks

vjohnson1-godaddy: By "multiple metric correction" do we mean false discovery rate adjustment?

If so, I am _fairly_ confident the _type_ of metric won't be material to the adjustment (however open to designing around this just in case).

Perhaps we could represent the metrics as a dictionary where the keys are the type of metric and the values are lists of those metrics? We could also move the concern of instantiating the metrics outside of the registration methods.

That could then help us generalize some of the other code in this class while still ensuring we can do metric-type level manipulations in the future

```python3
def register_metric(
        metric: Metric
    ) -> None:
        self.metrics[type(metric)].append(metric)

def unregister_metric(
        metric: Metric
    ) -> None:
       self.metrics[type(metric)].remove(metric)

def get_overall_sample_size(self) -> float:
    metrics = reduce(operator.add, self.metrics.values())
    return max(map(self._get_single_sample_size, metrics))
```

tests/test_sample_size/test_sample_size_calculator.py
```diff
@@ -0,0 +1,216 @@
...
+            test_numerator_variance,
+            test_denominator_mean,
+            test_denominator_variance,
+            test_covariance,
+            test_mde,
+        )
+        self.assertEqual(len(calculator.ratio_metrics), 2)
+
+    @patch("statsmodels.stats.power.NormalIndPower.solve_power")
+    def test_get_single_sample_size_normal(self, mock_solve_power):
```
vjohnson1-godaddy, line 106: Do we have any appetite for parameterizing this test?

```python3
@parameterized([
    {"metric": BooleanMetric(...), "solver": NormalIndPower},
    {"metric": NumericMetric(...), "solver": TTestIndPower}
])
@patch("statsmodels.stats.power.")
def test_get_single_sample_size_normal(self, mock_power, metric, solver):
    ...
```

tests/test_sample_size/test_sample_size_calculator.py
```diff
@@ -0,0 +1,216 @@
...
+        self.assertEqual(sample_size, test_sample_size)
+        mock_solve_power.assert_called_once()
+        assert_equal(mock_solve_power.call_args[1]["effect_size"], 0.15811388300841897)
+        assert_equal(mock_solve_power.call_args[1]["alpha"], DEFAULT_ALPHA)
+        assert_equal(mock_solve_power.call_args[1]["power"], DEFAULT_POWER)
+        assert_equal(mock_solve_power.call_args[1]["ratio"], 1)
+        assert_equal(mock_solve_power.call_args[1]["alternative"], "two-sided")
+
+    @patch("sample_size.sample_size_calculator.SampleSizeCalculator._get_single_sample_size")
+    def test_get_overall_sample_size_bool(self, mock_get_single_sample_size):
```
vjohnson1-godaddy, line 150: Similar question about parameterization here

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,81 @@
...
+                effect_size=effect_size,
+                alpha=self.alpha,
+                power=self.power,
+                ratio=1,
+                alternative="two-sided",
+            )
+        )
+        return sample_size
+
+    def get_overall_sample_size(self) -> float:
```
satish-ravi, line 67: should this just be `get_sample_size`? or do we plan to expose multiple methods, one for overall, one for individual?

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __init__(self, alpha: float = DEFAULT_ALPHA, power: float = DEFAULT_POWER):
+        self.alpha = alpha
+        self.power = power
+        # Consider having a self.metrics to hold all metric types
+        self.boolean_metrics: List[BooleanMetric] = []
+        self.numeric_metrics: List[NumericMetric] = []
+        self.ratio_metrics: List[RatioMetric] = []
+
+    def register_bool_metric(self, probability: float, mde: float) -> None:
```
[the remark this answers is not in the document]
wendychiang: why will we want to do that? I feel like that is out of scope considering how we'll use this package (at least not a requirement for now)

tests/test_sample_size/test_variables.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def test_boolean_metric_get_probability_too_large(self):
+        test_probability = 5
+
+        with self.assertRaises(Exception) as context:
+            BooleanMetric._get_probability(test_probability)
+            self.assertEqual(
+                context.exception,
+                Exception("Error: Please provide a float between 0 and 1 for probability."),
+            )
```
wendychiang, line 51: you need to move assertEqual outside of the `with` otherwise it doesn't work (it will just pass even when the values are not equal)

and looks like it can't compare Exception directly, maybe something like
```python
        self.assertEqual(
            str(context.exception),
            "Error: Please provide a float between 0 and 1 for probability.",
        )
```

tests/test_sample_size/test_variables.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def test_boolean_metric_get_probability_too_small(self):
+        test_probability = -0.1
+
+        with self.assertRaises(Exception) as context:
+            BooleanMetric._get_probability(test_probability)
+            self.assertEqual(
+                context.exception,
+                Exception("Error: Please provide a float between 0 and 1 for probability."),
+            )
```
wendychiang, line 61: same as previous comment

tests/test_sample_size/test_variables.py
```diff
@@ -0,0 +1,121 @@
...
+from sample_size.metrics import BooleanMetric
+from sample_size.metrics import NumericMetric
+from sample_size.metrics import RatioMetric
+
+
+class BooleanMetricTestCase(unittest.TestCase):
+    def setUp(self):
+        self.DEFAULT_MDE = 0.01
+        self.DEFAULT_PROBABILITY = 0.05
+        self.DEFAULT_MOCK_VALUE = 99
```
wendychiang, line 16:
```suggestion
        self.DEFAULT_VARIANCE = 99
```
? since it is mocked, so it's fine if the value is not the true "variance" I think

tests/test_sample_size/test_variables.py
```diff
@@ -0,0 +1,121 @@
+import unittest
```
wendychiang, line 1: naming nit: rename this file to test_metrics.py so it is consistent with the actual file

tests/test_sample_size/test_variables.py
```diff
@@ -0,0 +1,121 @@
...
+
+class RatioMetricTestCase(unittest.TestCase):
+    def setUp(self):
+        self.DEFAULT_MDE = 5
+        self.DEFAULT_NUMERATOR_MEAN = 2000
+        self.DEFAULT_NUMERATOR_VARIANCE = 100000
+        self.DEFAULT_DENOMINATOR_MEAN = 200
+        self.DEFAULT_DENOMINATOR_VARIANCE = 2000
+        self.DEFAULT_COVARIANCE = 5000
+        self.DEFAULT_MOCK_VALUE = 99
```
wendychiang, line 86:
```suggestion
        self.DEFAULT_MOCK_VARIANCE = 99
```

sample_size/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+from statsmodels.stats.power import TTestIndPower
+
+
+class BooleanMetric:
+    def __init__(
+        self,
+        probability: float,
+        mde: float,
+    ):
+        self.probability = self._get_probability(probability)
```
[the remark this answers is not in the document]
chuxiaofdn: done

sample_size/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _get_variance(self) -> float:
+        return self.probability * (1 - self.probability)
+
+    @staticmethod
+    def _get_probability(probability: float) -> float:
+        if 0 <= probability <= 1:
+            return probability
+        else:
+            raise Exception("Error: Please provide a float between 0 and 1 for probability.")
```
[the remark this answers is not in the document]
chuxiaofdn: done

sample_size/sample_size_calculator.py
chuxiaofdn: done

tests/test_sample_size/test_sample_size_calculator.py
```diff
@@ -0,0 +1,216 @@
+import unittest
```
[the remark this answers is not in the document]
chuxiaofdn: done

sample_size/sample_size_calculator.py
chuxiaofdn: I wonder if Boolean metric require TTestIndPower when we support multiple metrics per calculator. We are going to modify this method anyway after we know the correction implementation we are going to use. I will keep this comment unresolved as a reference

tests/test_sample_size/test_sample_size_calculator.py
chuxiaofdn: We could do this if we have more types in the future.

tests/test_sample_size/test_sample_size_calculator.py
chuxiaofdn: We could do this if we have more types in the future.

sample_size/sample_size_calculator.py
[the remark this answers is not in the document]
chuxiaofdn: Agree with @wendychiang. Unregister has no use case inside the class, and users can have a clear list of metrics before registering them.

sample_size/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+class BooleanMetric:
+    def __init__(
+        self,
+        probability: float,
+        mde: float,
+    ):
+        self.probability = self._get_probability(probability)
+        self.variance = self._get_variance()
+        self.mde = mde
+        self.default_power_analysis_type = NormalIndPower
```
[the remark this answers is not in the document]
chuxiaofdn: I update the metrics to store an instance of the power solver. Would a base class be necessary here?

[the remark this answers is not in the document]
wendychiang: I will +1 on adding a base class to connect these three classes for their common pattern, at least to define and make sure the common variable names (`self.variance`, `self.mde`, `self.default_power_analysis_type`) are consistent. Since they are used by calculator generally, with base class defined we can avoid mismatch variable names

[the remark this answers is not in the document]
chuxiaofdn: added

tests/sample_size/test_sample_size_calculator.py
```diff
@@ -0,0 +1,218 @@
...
+        mock_solve_power.assert_called_once_with(
+            effect_size=0.15811388300841897,
+            alpha=DEFAULT_ALPHA,
+            power=DEFAULT_POWER,
+            ratio=1,
+            alternative="two-sided",
+        )
+
+    @patch("sample_size.sample_size_calculator.SampleSizeCalculator._get_single_sample_size")
+    def test_get_overall_sample_size_bool(self, mock_get_single_sample_size):
```
vjohnson1-godaddy, line 152: nit:[ add comment reminding future contributors to parameterize these](https://github.com/godaddy/sample-size/pull/4#discussion_r761609769)

vjohnson1-godaddy (approved):

sample_size/metrics.py
```diff
@@ -0,0 +1,92 @@
+from typing import Union
+
+from statsmodels.stats.power import NormalIndPower
+from statsmodels.stats.power import TTestIndPower
+
+
+class BaseMetric:
```
wendychiang, line 7: how about adding a constructor and set mde because they are the same across three types, and we can potentially also set `variance` and `default_power_analysis_instance` too
```python
class BaseMetric(object):

    __metaclass__ = ABCMeta

    def __init__(self, mde: float):
        self.mde = mde
    
    @property
    @abstractmethod
    def default_power_analysis_instance(self) -> Union[NormalIndPower, TTestIndPower]:
        raise NotImplementedError

    @property
    @abstractmethod
    def variance(self) -> float:
        raise NotImplementedError
    
    @staticmethod
    def check_positive(number: float, name: str) -> float:
        if number < 0:
            raise Exception(f"Please provide a positive number for {name}.")
        return number
```
And in each metric, we can do
```python
class NumericMetric(BaseMetric):
    def __init__(
        self,
        variance: float,
        mde: float,
    ):
        super(NumericMetric, self).__init__(mde)
        self._variance = self.check_positive(variance, "variance")

    @property
    def variance(self) -> float:
        return self._variance
    
    @property
    def default_power_analysis_instance(self) -> NormalIndPower:
        return TTestIndPower()
```

sample_size/metrics.py
```diff
@@ -0,0 +1,92 @@
...
+
+    variance: float
+    mde: float
+    default_power_analysis_instance: Union[NormalIndPower, TTestIndPower]
+
+    @staticmethod
+    def check_positive(number: float, name: str) -> None:
```
wendychiang, line 14: maybe can return the number if it is positive? then it becomes one line only
```python
self.numerator_variance = self.check_positive(numerator_variance)
```

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,81 @@
...
+from sample_size.metrics import BooleanMetric
+from sample_size.metrics import NumericMetric
+from sample_size.metrics import RatioMetric
+
+DEFAULT_ALPHA = 0.05
+DEFAULT_POWER = 0.8
+
+
+class SampleSizeCalculator:
+    """This class is to calculate sample size based on variable type
```
wendychiang, line 15:
```suggestion
    """
    This class is to calculate sample size based on variable type
```

sample_size/sample_size_calculator.py
```diff
@@ -0,0 +1,81 @@
...
+        denominator_variance: float,
+        covariance: float,
+        mde: float,
+    ) -> None:
+        metric = RatioMetric(
+            numerator_mean, numerator_variance, denominator_mean, denominator_variance, covariance, mde
+        )
+        self.ratio_metrics.append(metric)
+
+    def _get_single_sample_size(self, metric: Union[BooleanMetric, NumericMetric, RatioMetric]) -> float:
```
wendychiang, line 53:
```suggestion
    def _get_single_sample_size(self, metric: BaseMetric) -> float:
```
probably can just do BaseMetric now?

sample_size/metrics.py
```diff
@@ -0,0 +1,124 @@
...
+        numerator_variance: float,
+        denominator_mean: float,
+        denominator_variance: float,
+        covariance: float,
+        mde: float,
+    ):
+        super(RatioMetric, self).__init__(mde)
+        self.numerator_mean = numerator_mean
+        self.check_positive(numerator_variance, "variance")
+        self.numerator_variance = numerator_variance
```
wendychiang, line 105:
```suggestion
        self.numerator_variance = self.check_positive(numerator_variance, "variance")
```
can simplify now, same for denominator_variance

sample_size/metrics.py
```diff
@@ -0,0 +1,124 @@
...
+        self.denominator_mean = denominator_mean
+        self.check_positive(denominator_variance, "variance")
```
wendychiang, line 107:
```suggestion
        self.check_positive(denominator_variance, "denominator_variance")
```
nit, what do you think?

wendychiang (approved):
+1 with couple small comments

PR #6, opened 2021-11-30
add run scripts and utils to get user input

chuxiaofdn:
The prior [PR](https://github.com/godaddy/sample-size/pull/2) has been split into two PRs for better review. 
This PR contains the run script, utils, and tests.

* [ ] Tests (if applicable)

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)


scripts/utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+    parameters = {}
+
+    for parameter_name, parameter_definition in parameter_definitions.items():
+        parameters[parameter_name] = get_input(parameter_definition)
+
+    return parameters
+
+
+def register_metric(metric_type: str, calculator: SampleSizeCalculator) -> None:
```
wendychiang, line 65: this function is actually doing not just register but also getting inputs, how about either rename the function, or refactor so there is another function taking care the input, and this one will just do register?
something like
```python
def get_metric_metadata_from_input():
    metric_type = get_metric_type_from_input().lower()
    metric_metadata = get_variable_parameters(METRIC_PARAMETERS[metric_type])
    metric_metadata['mde'] = get_mde(metric_type)
    return metric_type, metric_metadata
    
    
def register_metric(metric_type: str, metric_metadata, calculator: SampleSizeCalculator) -> None:
    VAR_REGISTER_FUNC_MAP = {
        'boolean': 'register_bool_metric',
        'numeric': 'register_numeric_metric',
        'ratio': 'register_ratio_metric'
    }
    fregister_func = getattr(calculator, VAR_REGISTER_FUNC_MAP[metric_type])
    fregister_func(**metric_metadata)
```

Then the usage in main() will be
```python
metric_type, metric_metadata = get_metric_metadata_from_input()
register_metric(metric_type, metric_metadata, calculator)
```

scripts/sample_size_run.py
```diff
@@ -0,0 +1,34 @@
...
+    # Get alpha for power analysis
+    alpha = get_alpha()
+    if alpha:
+        calculator = SampleSizeCalculator(alpha)
+    else:
+        calculator = SampleSizeCalculator()
+
+    # register metric
+    metric_type = get_variable_from_input()
+    register_metric(metric_type, calculator)
```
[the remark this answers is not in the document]
chuxiaofdn: Good point.Will add this.

scripts/sample_size_run.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    from sample_size.sample_size_calculator import SampleSizeCalculator
+    from scripts.utils import get_alpha
+    from scripts.utils import get_variable_from_input
+    from scripts.utils import register_metric
+
+    # Get alpha for power analysis
+    alpha = get_alpha()
+    if alpha:
+        calculator = SampleSizeCalculator(alpha)
```
wendychiang, line 20: does that mean `power` is fixed from the script that we don't want users to input it?

tests/test_scripts/test_sample_size_run.py
```diff
@@ -0,0 +1,108 @@
+import unittest
```
wendychiang, line 1: naming nit: rename directory to be under tests/scripts

scripts/utils.py
```diff
@@ -0,0 +1,96 @@
...
+def is_float(value: str) -> float:
+    try:
+        float(value)
+        return True
+    except ValueError:
+        return False
+
+
+def get_raw_input(text: str) -> str:
+    return input(text)
```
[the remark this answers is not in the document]
wendychiang: just do `@patch("scripts.utils.input")` to mock input() , then we can remain 100% coverage 🙂

scripts/sample_size_run.py
chuxiaofdn: Internally we supports configurable alpha for now, so the script to configure power is not included here. IMO, most of the users will not update statistical power, asking them to input the power may add inconvenience.

chuxiaofdn: note added

tarkatronic:
You might also want to consider adding a `[tool.poetry.scripts]` section to the `pyproject.toml` for your scripts. For example: https://github.com/godaddy/tartufo/blob/main/pyproject.toml#L36-L37

This way, when people install the project, the script/command is automatically available to them. So for example you could have:

```toml
[tool.poetry.scripts]
run-sample-size = "scripts.sample_size_run:main"
```
And then users would have a `run-sample-size` command available to them.

One other big thing to keep in mind is the actual paths you are using. Since `scripts` is a top-level module here, that means that it will get bundled by `poetry` when you build a wheel with `poetry build`. Then, when an end-user installs that wheel, they will then have a module named `scripts` in their Python installation. Vitally, this will not be scoped under `sample_size`, but rather global. This can be dangerous, as it might conflict with other projects/packages.

scripts/utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+def get_alpha() -> Union[float, None]:
+    alpha_default_check = get_raw_input("Do you want to use default alpha (0.05) for the power analysis? (y/n)")
+    if alpha_default_check.lower() == "n":
+        alpha = get_input("alpha")
+        if 0 <= alpha <= 0.3:
+            return alpha
+        else:
+            raise Exception("Error: Please provide a float between 0 and 0.3 for alpha.")
+    else:
```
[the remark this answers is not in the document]
wendychiang: how about making it into just one question? e.g
```
Enter the alpha between (between 0 and 0.3 inclusively) or press enter to use default alpha=0.05
```
then checking input equals empty string to set alpha = 0.05, otherwise throw error?

wendychiang:
Maybe can include some instructions on how to use this script somewhere (we can add more details later after we publish the package) then it is clear about how to use this script

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+        self.assertEqual(
+            str(context.exception),
+            f"Error: Please enter a float for the {self.TEST_STR}.",
+        )
+
+    @patch("sample_size.scripts.utils.get_float")
+    @patch("sample_size.scripts.utils.input")
+    def test_get_alpha(self, mock_input, mock_get_float):
+        test_input_float = 0.01
+        mock_input.return_value = "n"
```
wendychiang, line 53:
```suggestion
        mock_input.return_value = "0.01"
```
otherwise, input of `n` should raise exception

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+
+        with patch("sys.stdout", new=StringIO()) as fakeOutput:
+            alpha = utils.get_alpha()
+            self.assertEqual(
+                fakeOutput.getvalue().strip(),
+                f"Using alpha ({alpha}) and default power (0.8)...",
+            )
+
+        self.assertEqual(alpha, test_input_float)
+        mock_input.called_once_with(
```
wendychiang, line 64:
```suggestion
        mock_input.assert_called_once_with(
```

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+            "Enter the alpha between (between 0 and 0.3 inclusively) " "or press Enter to use default alpha=0.05: "
+        )
+        mock_get_float.called_once_with(test_input_float, "alpha")
```
wendychiang, line 67:
```suggestion
        mock_get_float.assert_called_once_with("0.01", "alpha")
```
it is not doing assertion now, also need to fix the values to be "n" (base on your current code) or "0.01" based on my suggestion

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @patch("sample_size.scripts.utils.input")
+    def test_get_alpha_default(self, mock_input):
+        mock_input.return_value = " "
+
```
wendychiang, line 72: can we also also assert get_float not called?

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+        self.assertEqual(
+            context.exception.args[0],
+            "Error: Please provide a float between 0 and 0.3 for alpha.",
+        )
+
+    @patch("sample_size.scripts.utils.get_float")
+    @patch("sample_size.scripts.utils.input")
+    def test_get_alpha_too_small(self, mock_input, mock_get_float):
+        test_input_float = -0.1
+        mock_input.return_value = "n"
```
wendychiang, line 101:
```suggestion
        mock_input.return_value = "-0.1"
```

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+        mock_input.return_value = test_mde
+        mock_get_float.return_value = test_mde
+
+        mde = utils.get_mde(test_metric_type)
+
+        self.assertEqual(mde, test_mde)
+        mock_input.called_once_with(
+            f"Enter the absolute minimum detectable effect for this {test_metric_type} \n"
+            f"MDE: targeted treatment metric value minus the baseline value"
+        )
```
wendychiang, line 126:
```suggestion
        mock_input.assert_called_once_with(
            f"Enter the absolute minimum detectable effect for this {test_metric_type} \n"
            f"MDE: targeted treatment metric value minus the baseline value: "
        )
```

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+        mock_get_float.return_value = test_input_float
+        test_parameter_definitions = {
+            "test": "test test",
+            "case": "case case",
+        }
+
+        result = utils.get_variable_parameters(test_parameter_definitions)
+
+        self.assertEqual(mock_input.call_count, 2)
+        assert_equal(mock_input.call_args_list[0][0][0], "Enter the test test: ")
```
wendychiang, line 179: why to use assert_equal here? self.assertEqual should just work?
```suggestion
        self.assertEqual(mock_input.call_args_list[0][0][0], "Enter the test test: ")
```

wendychiang (commented):
bumping some comments that are probably missed to be updated

sample_size/scripts/sample_size_run.py
```diff
@@ -0,0 +1,38 @@
...
+        2. the calculator supports single metric per calculator for now
+    """
+    from sample_size.sample_size_calculator import SampleSizeCalculator
+    from sample_size.scripts.utils import get_alpha
+    from sample_size.scripts.utils import get_metric_metadata_from_input
+    from sample_size.scripts.utils import register_metric
+
+    try:
+        # Get alpha for power analysis
+        alpha = get_alpha()
```
vjohnson1-godaddy, line 22: Should we also be collecting power here?

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,274 @@
...
+            {
+                "test": test_input_float,
+                "case": test_input_float,
+            },
+        )
+
+    @patch("sample_size.scripts.utils.get_mde")
+    @patch("sample_size.scripts.utils.get_variable_parameters")
+    @patch("sample_size.scripts.utils.get_metric_type")
+    def test_get_metric_metadata_from_input(self, mock_get_metric_type, mock_get_variable_parameters, mock_get_mde):
```
vjohnson1-godaddy, line 204: I wonder if we could approach testing this differently? In particular, if we could parameterize input as a number of test cases and focus exclusively on comparing the output of `get_metric_metadata_from_input` against some ground truth? Testing whether `mock_get_mde` was called, for example, might be an implementation detail that's better assessed by what MDE we return from `get_metric_metadata_from_input`.

tests/sample_size/scripts/test_utils.py
```diff
@@ -0,0 +1,274 @@
...
+        test_parameter_definitions = {
+            "test": "test test",
+            "case": "case case",
+        }
+
+        result = utils.get_variable_parameters(test_parameter_definitions)
+
+        self.assertEqual(mock_input.call_count, 2)
+        self.assertEqual(mock_input.call_args_list[0][0][0], "Enter the test test: ")
+        self.assertEqual(mock_input.call_args_list[1][0][0], "Enter the case case: ")
```
vjohnson1-godaddy, line 187: Because the two test parameter definitions are virtually identical, perhaps we could only check that `mock_input` and `mock_get_float` are called the same number of times as `len(test_parameter_definitions)` and only assess the individual output of, say, `"test": "test test"`?

sample_size/scripts/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+def get_metric_metadata_from_input() -> Tuple[str, Dict[str, float]]:
+    metric_type = get_metric_type().lower()
+    metric_metadata = get_variable_parameters(METRIC_PARAMETERS[metric_type])
+    metric_metadata["mde"] = get_mde(metric_type)
+
+    return metric_type, metric_metadata
+
+
+def register_metric(
```
vjohnson1-godaddy, line 90: Would it make sense to make this an instance method of `SampleSizeCalculator`?

sample_size/scripts/sample_size_run.py
wendychiang: FYI an existing thread about this (already resolved so i can only paste screenshot here)
image

vjohnson1-godaddy: Ope, sorry I missed that!

sample_size/sample_size_calculator.py
```diff
@@ -80,3 +81,13 @@ def get_sample_size(self) -> float:
                 sample_size = self._get_single_sample_size(ratio_metric)
 
         return sample_size
+
+    def register_metric(self, metric_type: str, metric_metadata: Dict[str, float]) -> None:
+        VAR_REGISTER_FUNC_MAP = {
+            "boolean": "register_bool_metric",
+            "numeric": "register_numeric_metric",
+            "ratio": "register_ratio_metric",
```
wendychiang, line 89: when this function is used by the script only, hardcoded the string key here seems fine. but if now we move it here, i'm wondering if there is a way to let the input metric_type always synced. e.g if anytime we rename "boolean" to "bool" in line 87, all apps will break.

one way i have in mind is to generate a constant in this file
```python
class MetricType:
    BOOL = "boolean"
    NUMERIC = "numeric"
    RATIO = "ratio"
```
Then utils.py could create another constant like below used in [line 65](https://github.com/godaddy/sample-size/blob/063d2748808321080aa8b88c44c926c6f95c4ed7/sample_size/scripts/utils.py#L65) by import MetricType 
```
METRIC_TYPES = [MetricType.BOOL, MetricType.RATIO, MetricType.NUMERIC]
```

and any other app could also do something like `calculator.register_metric(MetricType.BOOL, ..)`

But I don't know if that really matters at all considering if we actually rename, we will publish a new version anyway and it is probably unlikely we will rename though

wendychiang (approved):
looks good to me with a comment

tests/sample_size/scripts/test_utils.py
chuxiaofdn: Can you be more specific on your suggestion? The functions like `get_mde` has their own separate tests. So they are tested for properly called here. Can you provide some examples?

README.md
```diff
@@ -2,6 +2,23 @@
 
 This python project is a helper package that uses power analysis to calculate required sample size for any experiment.
 
+# Get Started
```
satish-ravi, line 5: nit: 
```suggestion
## Get Started
```

README.md
```diff
@@ -2,6 +2,23 @@
...
+```buildoutcfg
+pip install sample-size
+run-sample-size
+```
+
+# Notes
```
satish-ravi, line 11: Should this be converted to a usage guide on how to pass params to the script?

sample_size/metrics.py
```diff
@@ -27,7 +27,7 @@ def variance(self) -> float:
     @staticmethod
     def check_positive(number: float, name: str) -> float:
         if number < 0:
-            raise Exception(f"Please provide a positive number for {name}.")
+            raise Exception(f"Error: Please provide a positive number for {name}.")
```
satish-ravi, line 30: Can we use a more specific Exception, like ValueError instead of the generic Exception

sample_size/scripts/utils.py
```diff
@@ -0,0 +1,86 @@
...
+        "numerator_mean": "mean of the baseline metric's numerator",
+        "numerator_variance": "variance of the baseline metric's numerator",
+        "denominator_mean": "mean of the baseline metric's denominator",
+        "denominator_variance": "variance of the baseline metric's denominator",
+        "covariance": "covariance between the baseline metric's numerator and denominator",
+    },
+}
+
+
+def is_float(value: str) -> float:
```
satish-ravi, line 19: this is returning bool right? Does mypy pass this check?

sample_size/scripts/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    except ValueError:
+        return False
+
+
+def get_float(input_str: str, input_name: str) -> float:
+    input_str = input_str.strip()
+    if is_float(input_str):
+        return float(input_str)
+    else:
+        raise Exception(f"Error: Please enter a float for the {input_name}.")
```
satish-ravi, line 32: raise more specific exceptions here and below

sample_size/scripts/utils.py
```diff
@@ -0,0 +1,86 @@
+from typing import Dict
```
satish-ravi, line 1: Wonder if this file can be me named better than a generic `utils.py`. This seems to be doing bulk of the input processing right?
Also, have we considered using argparse for input processing? Which is typically standard for CLI tools..

sample_size/scripts/sample_size_run.py
```diff
@@ -0,0 +1,37 @@
...
+        # register metric
+        metric_type, metric_metadata = get_metric_metadata()
+        calculator.register_metric(metric_type, metric_metadata)
+
+        # Get and print sample size based on variable and power analysis parameters
+        sample_size = calculator.get_sample_size()
+        print("\nSample size needed in each group: {:.3f}".format(sample_size))
+    except Exception as e:
+        print(f"Error! The calculator isn't able to calculate sample size due to \n{e}")
+        return
```
satish-ravi, line 33: is this return needed?

README.md
satish-ravi, line 11: Wonder if this can be more specific than just Notes. Also we should move this to one level down in markdown headers similar to the above comment.

chuxiaofdn: I'll update it to Script Constraints.

sample_size/scripts/utils.py
chuxiaofdn: yeah somehow it passed. will update

sample_size/scripts/utils.py
chuxiaofdn: will update to input_utils. I'll add a comment to consider using click.

sample_size/scripts/sample_size_run.py
chuxiaofdn: not needed. will remove

tests/sample_size/scripts/test_utils.py
vjohnson1-godaddy: I've implemented a sketch of what I was describing in my own branch. Could you give me the permissions to be able to push my branch up to origin so I can link it here?

wendychiang: @vjohnson1-godaddy you need to first be added to godaddy org to be the contributor, follow the instruction here https://github.secureserver.net/hivemind/docs/blob/master/onboarding/hivemind-developer-onboarding.md#opensource-contribution

vjohnson1-godaddy: While I wait for my PR to be a contributor to be approved, I can clarify what I was talking about and provide some examples.

- Within the test `test_get_metric_metadata_from_input`, as I understand it, what we're really concerned about is that, given a set of inputs, we get a particular `metric_type` and `metric_metadata`.  The particular way that we implement that process could change, but as long as the same inputs produce the same outputs, the function as a whole is still valid. In the case of `mock_get_mde.assert_called_once_with(test_metric_type_lower)`, I can imagine valid scenarios where we change the implementation but the functionality remains the same. Perhaps `get_mde` could be replace, or have its function signature changed. In any case, we would want the test to pass regardless of changes to `get_med` _so long as the output is the same_.
- In this test, we only look at `Boolean` metrics, but we have some other metrics that we're concerned about as well. We could parameterize some of the input values like `mde` and check that all of the outputs match what we expect. That's what I meant by parameterization. In the example below, I left the role of defining what our expected output should be to snapshot testing, but one could also add the expected output as an additional parameter in the list of test cases passed to `@parameterized.expand(`

```python3

    @parameterized.expand(
        [
            # metric_type_input, parameters, mde,
            ("Boolean", {"probability": 0.01}, 0.05),
            ("Numeric", {"variance": 1}, 1),
            (
                "Ratio",
                {
                    "numerator_mean": 1,
                    "numerator_variance": 2,
                    "denominator_mean": 3,
                    "denominator_variance": 4,
                    "covariance": 5,
                },
                0.05,
            ),
        ]
    )
    @patch("sample_size.scripts.utils.get_mde")
    @patch("sample_size.scripts.utils.get_metric_parameters")
    @patch("sample_size.scripts.utils.get_metric_type")
    def test_get_metric_metadata(
        self, metric_type_input, parameters_input, mde_input, mock_get_metric_type, mock_get_metric_parameters, mock_get_mde
    ):
        mock_get_metric_type.return_value = metric_type_input
        mock_get_metric_parameters.return_value = parameters_input
        mock_get_mde.return_value = mde_input

        metric_type, metric_metadata = utils.get_metric_metadata()

        self.assertEqual(metric_type, metric_type_input.lower())
        self.snapshot.assert_match(  # type: ignore[attr-defined]
            json.dumps(metric_metadata, indent=2), f"{metric_type_input}.json"
        )
```

chuxiaofdn: seems like the script is trying to make sure `METRIC_PARAMETERS` is not changed. I'm not sure if it is necessary, since it is a constant. In `test_get_metric_parameters`, we already test that we will get correct parameters when provide correct parameter definitions like `METRIC_PARAMETERS`.

vjohnson1-godaddy: Oh, I see, I see. Sounds reasonable!

vjohnson1-godaddy (approved):

PR #24, opened 2022-04-19
feat(multiple_testing_power): add functions required for average power calculator

xyzou685:

* [ ] Tests (if applicable)

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)


xyzou685:
This is a very rough draft containing functions proposed in [ML-4990](https://jira.godaddy.com/browse/ML-4990). Here are some questions open to discuss:
* Should we integrate this calculation with the metric classes defined in metrics.py?
* Currently the data is simulated for each variant and each decision metrics. Will it be better if we simulate all variants at once for each decision metric?

Other action items I will include in next commit:
* Add flexibility: separate different types of metrics into different function/ classes that requires their own sets of inputs/ properties
* Add delta methods into simulation process

vjohnson1-godaddy:
> Should we integrate this calculation with the metric classes defined in metrics.py?

Definitely. A given metric can have its own p-value generating function. Do you plan on moving `multiple_testing_sample_size_calculator.py` under `sample-size`?

---

> Will it be better if we simulate all variants at once for each decision metric?

I'm not sure if we want to introduce the concept of a decision metric into this project? In the spirit of a simpler design, perhaps the onus of handling different tiers of metrics should fall outside the calculator? In principle, from the side of the platform, we could call the calculator more than once to determine if metrics are adequately powered for different tiers.

---

> Are we adding constraints on input values and data types as what the calculator currently did, or should we add all constraint check in hivemind-ui?

In an ideal world it would probably make sense to set up a data structure (probaby a dataclass in `hivemind-dataclasses`) which encapsulates validation and which ensures interoperability between [`hivemind-python-api`](https://github.com/gdcorp-dna/hivemind-python-api) and `sample-size`. However, because one of those is closed source and the other is open, I think we might have to define the data structure in `sample-size` and `import` it in `hivemind-python-api`. Regardless, once we have some abstraction in place we can worry about what `hivemind-ui` sends `hivemind-python-api` when it comes time to actually add the calculator to the portal. To answer your question directly, the code for doing the constraint-check should probably live in `sample-size` as a dataclass and how `hivemind-ui` handles inputs and errors is an implementation detail for down the road.
 
---

Some high level comments:
- We don't need to simulate individual observations in the p-value generating function
   - A nice way to handle the p-value generating function might be for it to live in `SampleSizeCalculator` and have access to the registered metrics. Each metric could have its own (p-value generator)-generator, which would return a function which samples the distribution of the p-values (taking as argument whether the result is truly null or not). Thus for each simulation, we just take the list of registered metrics, sample a p-value for each metric in the list, and repeat.
- Might be a premature optimization, but we could take advantage of algorithms like [these](https://docs.scipy.org/doc/scipy/tutorial/optimize.html) to improve our search for the sample-size estimate that minimizes, for example, the squared-error between desired power and what `expected_average_power` returns

xyzou685:
> > Should we integrate this calculation with the metric classes defined in metrics.py?
> 
> Definitely. A given metric can have its own p-value generating function. Do you plan on moving `multiple_testing_sample_size_calculator.py` under `sample-size`?

Sure, I will add p-value generator to each metric class. I plan to finally dissect `multiple_testing_sample_size_calculator.py` by functions and add each function to different script, e.g. function `required_sample_size` will be added and called in `sample_size_run.py`

> > Will it be better if we simulate all variants at once for each decision metric?
> 
> I'm not sure if we want to introduce the concept of a decision metric into this project? In the spirit of a simpler design, perhaps the onus of handling different tiers of metrics should fall outside the calculator? In principle, from the side of the platform, we could call the calculator more than once to determine if metrics are adequately powered for different tiers.

That makes sense. I will make the calculator ask for input multiple times so the users can enter parameters for each variant for each metric

>  [`hivemind-python-api`](https://github.com/gdcorp-dna/hivemind-python-api) 

I got not found error for this repo. Do I need to be added for certain group to get the access to it?

chuxiaofdn:
> * Should we integrate this calculation with the metric classes defined in metrics.py?
I think this multiple testing correction deserves its own file. We would need to know the number and type of metrics registered in the calculator and then determine the power we should use to calculate sample size, no? 

> * Currently the data is simulated for each variant and each decision metrics. Will it be better if we simulate all variants at once for each decision metric?
We would assume all metrics registered together are the decision metrics people use to calculate sample size. I expect we should simulate all metrics at once or in parallel, whichever performs better.

> * Are we adding constraints on input values and data types as what the calculator currently did, or should we add all constraint check in hivemind-ui?

it makes more sense to add them here since the calculator is open-sourced. We would want it to be functional standalone.

sample_size/multiple_testing_funcs.py
```diff
@@ -0,0 +1,93 @@
...
+    upper = power_analysis.solve_power(
+        effect_size=d, alpha=self.alpha / m, power=self.power, ratio=1, alternative="two-sided"
+    )
+
+    # print(f'We look for the minimum required sample size in range [{int(lower)},{int(upper)}]')
+    for size in np.linspace(lower, upper, 10):
+        expected_power = self.expected_average_power(m, int(size))
+        if expected_power >= self.power:
+            break
+    return int(size)
```
vjohnson1-godaddy, line 29: You could turn this into a binary search


```python3
    return find_sample_size(lower, upper, metrics)

def find_sample_size(metrics, power, lower, upper, depth=0):
     if depth > MAX_RECURSION_DEPTH:
         raise RecursionError

    candidate = int((upper + lower) / 2)
    expected_power = expected_average_power(metrics, alpha, candidate)
   
     if np.isclose((power, expected_power), EPSILON):
        return candidate
   
     if expected_power > power:
         return find_sample_size(metrics, power, lower, candidate, depth + 1)
     else:
         return find_sample_size(metrics, power, candidate, upper, depth + 1)

def expected_average_power(metrics, alpha, sample_size):
     m = sum([m.variants - 1 for m in metrics])
     power = []
     for m1 in range(1, m+1):
         alts = np.array([True] * m1 + [False] * (m - m1))
          for _ in range(REPLICATIONS):
              true_alt = alts[np.random.permutation(m)]
              p_values = [m.generate_p_value(true_null[i]) for i, m in enumerate(metrics)]
              rejected = multipletests(p_values, alpha=alpha, method="fdr_bh")[0]
              power.append(np.dot(rejected, true_alt) / m1)
    
   return np.mean(power)
```

vjohnson1-godaddy (commented):
I think `_get_single_sample_size` might actually make more sense under `BaseMetric` with the form

```python3
    def single_sample_size(self, alpha: float, power: float) -> int:
        effect_size = self.mde / float(np.sqrt(self.variance))

        return int(
            self.default_power_analysis_instance.solve_power(
                effect_size=effect_size,
                alpha=self.alpha,
                power=self.power,
                ratio=1,
                alternative="two-sided",
            )
        )
```

The reason being is that the definition of `_get_single_sample_size` is really dependent on these other instance variables defined in `BaseMetric` (e.g. `default_power_analysis_instance`). What is nice about this approach is that you can describe 

```python3
    lower = power_analysis.solve_power(
        effect_size=d, alpha=self.alpha, power=self.power, ratio=1, alternative="two-sided"
    )
    upper = power_analysis.solve_power(
        effect_size=d, alpha=self.alpha / m, power=self.power, ratio=1, alternative="two-sided"
    )
```
as
```
    lower = min([m.single_sample_size(alpha, power) for m in metrics])
    upper = max([m.single_sample_size(alpha / m, power) for m in metrics])

```

sample_size/scripts/input_utils.py
```diff
@@ -86,3 +86,29 @@ def get_metric_metadata() -> Tuple[str, Dict[str, float]]:
...
+def get_variants() -> int:
+    print(f"")
+    number_of_variants = get_int(
+        input("Enter the number of cohorts for this test \n" "Control + number of treatments: "), "number of variants"
+    )
+    return number_of_variants
+
+
+def register_another_metric() -> bool:
+    register = input("Are you going to register another metric? (Y/N)")
```
dfu-godaddy, line 108:
```
register = input("Are you going to register another metric? (Y/N)").strip().lower()
if register=="y":
    return True
elif register=="n" or register=="":
    return Flase
```

sample_size/sample_size_calculator.py
```diff
@@ -1,15 +1,16 @@
...
 from typing import List
-
 import numpy as np
-
 from sample_size.metrics import BaseMetric
 from sample_size.metrics import BooleanMetric
 from sample_size.metrics import NumericMetric
 from sample_size.metrics import RatioMetric
 
+from multiple_testing_funcs import get_multiple_sample_size
```
dfu-godaddy, line 9: `from sample_size.multiple_testing_funcs import get_multiple_sample_size`

sample_size/scripts/sample_size_run.py
```diff
@@ -15,19 +15,32 @@ def main() -> None:
...
+        calculator = SampleSizeCalculator(alpha, variants)
 
-        # register metric
-        metric_type, metric_metadata = get_metric_metadata()
-        calculator.register_metric(metric_type, metric_metadata)
+        register = True
+        while register:
+            metric_type, metric_metadata = get_metric_metadata()
+            for v in range(int(variants) - 1):
+                print(f'Treatment {v + 1}')
```
dfu-godaddy, line 32: why are we printing out the treatment number?

sample_size/scripts/sample_size_run.py
```diff
@@ -15,19 +15,32 @@ def main() -> None:
...
         alpha = get_alpha()
-        calculator = SampleSizeCalculator(alpha)
+        variants = get_variants()
+        calculator = SampleSizeCalculator(alpha, variants)
 
-        # register metric
-        metric_type, metric_metadata = get_metric_metadata()
-        calculator.register_metric(metric_type, metric_metadata)
+        register = True
+        while register:
```
dfu-godaddy, line 29: can we move the codes for registration to `get_metric_metadata` under `input_utils.py`?
here can just be 
```
alpha = get_alpha()
variants = get_variants()
calculator = SampleSizeCalculator(alpha, variants)

metric_types, metric_metadatas = get_metric_metadata()
calculator.register_metric(metric_types, metric_metadatas)

# Get and print sample size based on variable and power analysis parameters
sample_size = calculator.get_sample_size()
```

dfu-godaddy: The current calculator gives me this error
image

sample_size/multiple_testing_funcs.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def generate_p_value(self, metric: BaseMetric, size: int) -> Tuple[List[float], List[float]]:
+    metric_type: str = type(metric).__name__
+    p_alt = np.zeros(self.rep)
+    p_null = stats.norm.rvs(0, 1, self.rep)
+    effect_size = metric.mde / float(np.sqrt(metric.variance))
+    z_alt = stats.t.rvs(df=size - 1, loc=effect_size, size=self.rep)
+
+    if metric_type == "BooleanMetric":
```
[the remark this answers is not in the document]
xyzou685: Oh I forgot to bring this up in spike check-in. Definitely will do the mapping, but before that I want to make sure we all agree to the distribution of test statistics `z_alt` for each metric type. We are using z test for boolean and ratio metrics, but if we sample `z_alt` from z distribution then sample size will not be involved. If we are using different `z_alt` then `z_alt` will also need to be mapped

sample_size/sample_size_calculator.py
```diff
@@ -22,13 +23,15 @@ class SampleSizeCalculator:
 
     """
 
-    def __init__(self, alpha: float = DEFAULT_ALPHA, power: float = DEFAULT_POWER):
+    def __init__(self, alpha: float = DEFAULT_ALPHA, variants: int = DEFAULT_VARIANTS, power: float = DEFAULT_POWER):
         self.alpha = alpha
         self.power = power
+        self.variants = variants
```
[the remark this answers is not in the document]
xyzou685: I already assume that way as I set `DEFAULT_VARIANTS = 2` and ask users to input variants including control in `input_utils.py`

sample_size/multiple_testing_funcs.py
```diff
@@ -0,0 +1,93 @@
...
+    m = (len(self.boolean_metrics) + len(self.numeric_metrics) + len(self.ratio_metrics)) * self.variants
+    # calculate required sample size based on minimum standardized effect size since it requires maximum sample size
+    lower = power_analysis.solve_power(
+        effect_size=d, alpha=self.alpha, power=self.power, ratio=1, alternative="two-sided"
+    )
+    upper = power_analysis.solve_power(
+        effect_size=d, alpha=self.alpha / m, power=self.power, ratio=1, alternative="two-sided"
+    )
+
+    # print(f'We look for the minimum required sample size in range [{int(lower)},{int(upper)}]')
```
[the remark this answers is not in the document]
xyzou685: I kept that since I thought it can be one of the output provided to the users, but now it seems redundant. Will remove it

sample_size/sample_size_calculator.py
```diff
@@ -22,13 +23,15 @@ class SampleSizeCalculator:
...
         # Consider having a self.metrics to hold all metric types
         self.boolean_metrics: List[BooleanMetric] = []
         self.numeric_metrics: List[NumericMetric] = []
         self.ratio_metrics: List[RatioMetric] = []
+        self.rep: int = 100
 
     def register_bool_metric(self, probability: float, mde: float) -> None:
```
dfu-godaddy, line 36: Can we have a main `register_metrics(self, metrics: list[Dict[str, Any]])` in `sample_size_calculator` so that hivemind api can just pass in a list of metrics all together. `register_metrics` should take a list of metrics, the metrics should be dictionaries.

```
metrics = [
        {
            "metric_type": "boolean",
            "metric_metadata": {
                "probability": 0.05,
                "mde": 0.02,
            },
        },
    ],
```

sample_size/sample_size_calculator.py
```diff
@@ -22,13 +23,15 @@ class SampleSizeCalculator:
...
-    def __init__(self, alpha: float = DEFAULT_ALPHA, power: float = DEFAULT_POWER):
+    def __init__(self, alpha: float = DEFAULT_ALPHA, variants: int = DEFAULT_VARIANTS, power: float = DEFAULT_POWER):
         self.alpha = alpha
         self.power = power
+        self.variants = variants
         # Consider having a self.metrics to hold all metric types
         self.boolean_metrics: List[BooleanMetric] = []
         self.numeric_metrics: List[NumericMetric] = []
         self.ratio_metrics: List[RatioMetric] = []
+        self.rep: int = 100
```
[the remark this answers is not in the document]
pchiang-godaddy: what is rep? can we name it more descriptive?

sample_size/scripts/sample_size_run.py
```diff
@@ -15,19 +15,32 @@ def main() -> None:
...
     from sample_size.scripts.input_utils import get_metric_metadata
+    from sample_size.scripts.input_utils import get_mde
+    from sample_size.scripts.input_utils import get_variants
+    from sample_size.scripts.input_utils import register_another_metric
 
     try:
         # Get alpha for power analysis
         alpha = get_alpha()
-        calculator = SampleSizeCalculator(alpha)
+        variants = get_variants()
```
pchiang-godaddy, line 25: can we check if we need to update readme? https://github.com/godaddy/sample-size#script-constraints
(also I suggest having two separate PRs for easier review)

pchiang-godaddy (commented):
probably can make two PRs for easier review. one for updating the class `SampleSizeCalculator`, one for updating the script based on the previous branch

sample_size/scripts/sample_size_run.py
xyzou685: Oh it's supposed to be printed only when variants>2. For each treatment we are only asking for MDE so users don't need to enter other metadata again. For example if you have 3 treatments, the calculator will ask the following
```
Metric type
baseline
MDE

Treatment 2
MDE

Treatment 3
MDE

Are you going to register another metric? (Y/N)
```

To fix this I've changed the code to 
```
  for v in range(int(variants) - 1):
      if v > 0:
          print(f'Treatment {v + 1}')
          metric_metadata["mde"] = get_mde(metric_type)
```

dfu-godaddy: hmm...given a metric,  asking for different MDEs per variant may confuse people? It is hard for them to determine one MDE already. Can we just require one MDE per metric? Thoughts? @pchiang-godaddy @vjohnson1-godaddy

vjohnson1-godaddy: I agree. From a different perspective, if an experimenter is having trouble determining a MDE, perhaps they could enter a range? From our end, it doesn't add much complexity to support a uniform prior on the parameters we're using to generate the p-values. Might be a good V2 / V3 discussion

xyzou685: Do the users usually apply same MDE for all treatment variants? If that's the case then we will use one MDE for treatments. Same for the MDE range - if determining a specific MDE is often a blocker for the users, I'm totally fine with adding this feature.

xyzou685:
Updated the PR with the following major changes and several minor changes:
* Replace register_boolean/numeric/ratio_metric function by one register function, and self.metrics contains all metrics regardless of their type
* Make Multiple testing a class
* Replace find_sample_size’s iteration by binary search
* Move get_single_sample_size to BaseMetric

sample_size/metrics.py
```diff
@@ -1,19 +1,32 @@
...
 
 class BaseMetric:
-
     __metaclass__ = ABCMeta
     mde: float
 
     def __init__(self, mde: float):
         self.mde = mde
 
+    def single_sample_size(self, alpha: float, power: float) -> int:
```
dfu-godaddy, line 17: this is duplicate with `SampleSizeCalculator._get_single_sample_size`. We should keep sample size calculation in the SampleSizeCalculator

vjohnson1-godaddy: @dfu-godaddy I think this is a direct response to [this comment I made](https://github.com/godaddy/sample-size/pull/24#pullrequestreview-959656809). However, I think there might have been some miscommunication on my part. 

What I was suggesting was something more like

```python3
class BaseMetric:
    def single_sample_size(self) -> float:
        raise NotImplementedError

class Numeric(BaseMetric):
    def single_sample_size(self):
        effect_size = self.mde / float(sqrt(self.variance))

         return int(
             stats.ttest.solve_power(
                 effect_size=effect_size,
                 alpha=alpha,
                 power=power,
                 ratio=1,
                 alternative="two-sided",
             )
         )
```
```python3
class SampleSizeCalculator:
    def get_sample_size(self) -> float:
        return self.get_multiple_sample_size()

    def get_multiple_sample_size(self) -> int:
        if len(self.metrics) < 2:
            return self.metrics[0].single_sample_size()
        lower = max([metric.single_sample_size(self.alpha, self.power) for metric in self.metrics])
        upper = max([metric.single_sample_size(self.alpha / self.m, self.power) for metric in self.metrics])

        return self._find_sample_size(lower, upper)
```

What @dfu-godaddy is saying about duplication is true, and we should avoid it. In terms of keeping power-calculation encapsulated in one class, that could also work
```python3
class BaseMetric:
    # no single_sample_size
       

class Numeric(BaseMetric):
    # no single_sample_size
```

```python3
class SampleSizeCalculator:
    def get_sample_size(self) -> int:
        return self.get_multiple_sample_size()

    def _single_sample_size(self, metric) -> int:
       if type(metric) == Numeric:
           return stats.ttest.solve_power(
              ...
           )
       elif type(metric) == Ratio:
           return ...
 

    def get_multiple_sample_size(self) -> int:
        if len(self.metrics) < 2:
            return self.metrics[0].single_sample_size()

        lower = max([metric.single_sample_size(self.alpha, self.power) for metric in self.metrics])
        upper = max([metric.single_sample_size(self.alpha / self.m, self.power) for metric in self.metrics])

        return self._find_sample_size(lower, upper)
```

I can live with the the second implementation, but generally if you're using a type-based switch-case statement, it's better to make the functionality of that switch-case _a property of the type_

sample_size/multiple_testing_funcs.py
```diff
@@ -1,93 +1,80 @@
...
+            return stats.uniform.rvs(0, 1)
+
+        if metric_type in ["BooleanMetric", "RatioMetric"]:
+            z_alt = stats.t.rvs(df=size - 1, loc=effect_size, size=self.REPLICATION)
+            return 2 * stats.norm.sf(np.abs(z_alt))
+
+        elif metric_type == "NumericMetric":
+            nc = np.sqrt(size / 2) * metric.mde / metric.variance
+            t_alt = stats.nct.rvs(nc=nc, df=2 * (size - 1), size=self.REPLICATION)
+            return 2 * stats.t.sf(np.abs(t_alt))
```
vjohnson1-godaddy, line 80: Similar to the comment above. Switch-case statements based on the type of the metric are generally something better handled by the types themselves

sample_size/sample_size_calculator.py
```diff
@@ -70,46 +50,10 @@ def _get_single_sample_size(self, metric: BaseMetric) -> float:
...
-                        min_standardized_effect_size = numeric_metric.mde / np.sqrt(numeric_metric.variance)
-                        min_metric = numeric_metric
-            if self.ratio_metrics:
-                for ratio_metric in self.ratio_metrics:
-                    if min_standardized_effect_size > ratio_metric.mde / np.sqrt(ratio_metric.variance):
-                        min_standardized_effect_size = ratio_metric.mde / np.sqrt(ratio_metric.variance)
-                        min_metric = ratio_metric
-            power_analysis = min_metric.default_power_analysis_instance
-            sample_size = get_multiple_sample_size(min_standardized_effect_size, power_analysis)
+            sample_size = MultipleTesting(self.metrics, self.variants, self.alpha,
```
vjohnson1-godaddy, line 56: All 4 of the arguments you're passing to `MultipleTesting` are from the state of the `SampleSizeCalculator` instance. I think encapsulating the logic around mutliple testing into its own class is fine, but a better way of doing that might be the use of a "mix-in" class

```python
from .multiple_testing_funcs import MultipleTestingMixin

class SampleSizeCalculator(MultipleTestingMixin):
     def get_sample_size(self) -> float:
        return self.get_multiple_sample_size()

class MultipleTestingMixin:
    # note no __init__
    def get_multiple_sample_size(self) -> int:
        if len(self.metrics) < 2:
            return self.metrics[0].single_sample_size()
        lower = max([metric.single_sample_size(self.alpha, self.power) for metric in self.metrics])
        upper = max([metric.single_sample_size(self.alpha / self.m, self.power) for metric in self.metrics])

        return self._find_sample_size(lower, upper)

    def _find_sample_size(self, lower: float, upper: float, depth=0):
        MAX_RECURSION_DEPTH = 20
        EPSILON = 0.025

        if depth > MAX_RECURSION_DEPTH:
            raise RecursionError

        candidate = int((upper + lower) / 2)
        expected_power = self._expected_average_power(candidate)

        if np.isclose((self.power, expected_power), EPSILON) or np.isclose((candidate, upper), EPSILON):
            return candidate

        if expected_power > self.power:
            return self._find_sample_size(lower, candidate, depth + 1)
        else:
            return self._find_sample_size(candidate, upper, depth + 1)

    def _expected_average_power(self, sample_size: int):
        power = []
        for m1 in range(1, self.m + 1):
            alts = np.array([True] * m1 + [False] * (self.m - m1))
            for _ in range(self.REPLICATION):
                true_alt = alts[np.random.permutation(self.m)]
                p_values = [self._generate_p_value(m, true_alt[i], sample_size) for i, m in enumerate(self.metrics)]
                rejected = multipletests(p_values, alpha=self.alpha, method="fdr_bh")[0]
                power.append(np.dot(rejected, true_alt) / m1)

        return np.mean(power)
```

sample_size/metrics.py
xyzou685: > @dfu-godaddy I think this is a direct response to [this comment I made](https://github.com/godaddy/sample-size/pull/24#pullrequestreview-959656809). However, I think there might have been some miscommunication on my part.
> 
> What I was suggesting was something more like
> 
> ```python
> class BaseMetric:
>     def single_sample_size(self) -> float:
>         raise NotImplementedError
> 
> class Numeric(BaseMetric):
>     def single_sample_size(self):
>         effect_size = self.mde / float(sqrt(self.variance))
> 
>          return int(
>              stats.ttest.solve_power(
>                  effect_size=effect_size,
>                  alpha=alpha,
>                  power=power,
>                  ratio=1,
>                  alternative="two-sided",
>              )
>          )
> ```
> 
>  ```python
> class SampleSizeCalculator:
>     def get_sample_size(self) -> float:
>         return self.get_multiple_sample_size()
> 
>     def get_multiple_sample_size(self) -> int:
>         if len(self.metrics) < 2:
>             return self.metrics[0].single_sample_size()
>         lower = max([metric.single_sample_size(self.alpha, self.power) for metric in self.metrics])
>         upper = max([metric.single_sample_size(self.alpha / self.m, self.power) for metric in self.metrics])
> 
>         return self._find_sample_size(lower, upper)
> ```
> 
> What @dfu-godaddy is saying about duplication is true, and we should avoid it. In terms of keeping power-calculation encapsulated in one class, that could also work
> 
> ```python
> class BaseMetric:
>     # no single_sample_size
>        
> 
> class Numeric(BaseMetric):
>     # no single_sample_size
> ```
> 
>  ```python
> class SampleSizeCalculator:
>     def get_sample_size(self) -> int:
>         return self.get_multiple_sample_size()
> 
>     def _single_sample_size(self, metric) -> int:
>        if type(metric) == Numeric:
>            return stats.ttest.solve_power(
>               ...
>            )
>        elif type(metric) == Ratio:
>            return ...
>  
> 
>     def get_multiple_sample_size(self) -> int:
>         if len(self.metrics) < 2:
>             return self.metrics[0].single_sample_size()
> 
>         lower = max([metric.single_sample_size(self.alpha, self.power) for metric in self.metrics])
>         upper = max([metric.single_sample_size(self.alpha / self.m, self.power) for metric in self.metrics])
> 
>         return self._find_sample_size(lower, upper)
> ```
> 
> I can live with the the second implementation, but generally if you're using a type-based switch-case statement, it's better to make the functionality of that switch-case _a property of the type_

It seems that my last changes on this function was missed from the commit. I did delete get_single_sample_size from `sample_size_calculator.py` and make `get_sample_size` as follows:
```
def get_sample_size(self) -> float:
        if len(self.metrics) == 1 and self.variants == DEFAULT_VARIANTS:
            sample_size = self.metrics[0].get_single_sample_size(self.alpha, self.power)
```

Also why can't we make it a BaseMetric method instead of a property needed to be implemented by each metric class?

vjohnson1-godaddy: > Why can't we make it a BaseMetric method 

I suppose it comes down to personal preference. But, in general, I'm not sure the construct `default_power_analysis_instance` completely makes sense to me. Specifically its unclear how we would choose a sensible default / what situation would necessitate inferring a metric type. Defining the implementation in the base class means you have reason about inheritance and remember what's been defined and what's being overridden. In a sense, its a "composition-vs-inheritance" issue. In context the only reason to define some inherited power-analysis instance is to avoid redundancy between the binomial and numeric cases which both use a z-test, but the calculation of effect size and variance is different between those cases. The way I see it, for the 3 metrics we have, there are 3 distinct power calculations.

But at the end of the day, this is not a blocker.

PR #26, opened 2022-05-13
feat(multiple_testing_power): add p-values generator to BaseMetric

xyzou685:

* [x] Tests (if applicable)

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [x] Yes (backward compatible)
* [ ] No (breaking changes)


xyzou685:
- add `generate_p_value` to each metric class so that for each registered metric, we are able to get its p-value(s) based on `true_alt` whether alternative hypothesis is true, sample size and number of variants. The output is an array with length=variants-1
- since there is only one certain power analysis type for each metric class, update the naming of `default_power_analysis_instance` to `power_analysis_instance`
- add unit test in `test_metric.py`
- the merge of this PR does not affect user experience

sample_size/metrics.py
```diff
@@ -31,6 +34,10 @@ def check_positive(number: float, name: str) -> float:
         else:
             return number
 
+    @abstractmethod
+    def generate_p_value(self, true_alt: bool, size: int, variants: int) -> Any:
```
vjohnson1-godaddy, line 38: I can't remember: Did we say we were going to make `variants` an instance attribute of metrics, or repeat each metric in the list of metrics?

sample_size/metrics.py
```diff
@@ -60,6 +67,15 @@ def _check_probability(probability: float) -> float:
...
 
+    def generate_p_value(self, true_alt: bool, size: int, variants: int) -> Any:
+        if not true_alt:
+            return stats.uniform.rvs(0, 1, size=variants - 1)
+
+        else:
+            nc = np.sqrt(size / 2) * self.mde / self.variance
+            z_alt = stats.nct.rvs(nc=nc, df=2 * (size - 1), size=variants - 1)
+            return 2 * stats.norm.sf(np.abs(z_alt))
+
```
vjohnson1-godaddy, line 78: I think the boolean metrics use a z-test, not a t-test, right?

```suggestion
    def generate_p_value(self, true_alt: bool, size: int, variants: int, n: int) -> Any:
        if not true_alt:
            return stats.uniform.rvs(0, 1, size=variants - 1)

        else:
            mu = self.mde / np.sqrt(self.variance / n)
            z_alt = stats.norm.rvs(loc=mu, size=variants - 1)
            return 2 * stats.norm.sf(np.abs(z_alt))

```

tests/sample_size/test_metrics.py
```diff
@@ -76,6 +77,74 @@ def test_boolean_metric_get_probability_too_small(self):
...
+    @patch("scipy.stats.uniform.rvs")
+    @patch("scipy.stats.nct.rvs")
+    def test_boolean_metric_generate_p_value_two_variants_true_alt(self, mock_nct, mock_uniform, mock_variance):
+        true_alt = True
+        size = 200
+        variants = 2
+        z = 0
+        pval = 1
+        nc = np.sqrt(size / 2) * self.DEFAULT_MDE / self.DEFAULT_MOCK_VARIANCE
+        mock_variance.__get__ = MagicMock(return_value=self.DEFAULT_MOCK_VARIANCE)
```
vjohnson1-godaddy, line 105: You can just set this up with

```suggestion
        boolean = BooleanMetric(self.DEFAULT_PROBABILITY, self.DEFAULT_MDE)
        boolean.varaicne = self.DEFAULT_MOCK_VARIANCE
```

tests/sample_size/test_metrics.py
```diff
@@ -76,6 +77,74 @@ def test_boolean_metric_get_probability_too_small(self):
...
+        pval = np.ones(variants - 1)
+        nc = np.sqrt(size / 2) * self.DEFAULT_MDE / self.DEFAULT_MOCK_VARIANCE
+        mock_variance.__get__ = MagicMock(return_value=self.DEFAULT_MOCK_VARIANCE)
+        mock_nct.return_value = z
+
+        boolean = BooleanMetric(self.DEFAULT_PROBABILITY, self.DEFAULT_MDE)
+        p = boolean.generate_p_value(true_alt, size, variants)
+        mock_uniform.assert_not_called()
+        mock_nct.assert_called_once_with(nc=nc, df=2 * (size - 1), size=variants - 1)
+        self.assertTrue((p == pval).all())
```
vjohnson1-godaddy, line 146: Might be a good place to parameterize (same goes for other classes as well)

```suggestion
    @parameterized.expand(product([True, False], [2, 5], [200]))
    @patch("scipy.stats.uniform.rvs")
    @patch("scipy.stats.nct.rvs")
    @patch("scipy.stats.norm.sf")
    def test_generate_p_value(self, mock_sf, mock_nct, mock_uniform, is_null, variants, size):
        p_value_generator = mock_uniform if is_null else mock_sf
        pval = MagicMock()
        p_value_generator.return_value = pval
        
        boolean = BooleanMetric(self.DEFAULT_PROBABILITY, self.DEFAULT_MDE)
        boolean.variance = self.DEFAULT_MOCK_VARIANCE
        p = boolean.generate_p_value(true_alt, size, variants)
        
        if is_null:
            mock_uniform.assert_called_once_with(0, 1, size=variants - 1)
            mock_nct.assert_not_called()
        else:
             mock_uniform.assert_not_called()
             mock_nct.assert_called_once_with(nc=nc, df=2 * (size - 1), size=variants - 1)
        self.assertTrue((p == pval).all())
```

sample_size/metrics.py
xyzou685: We were certain about not repeating each metric for `variants-1` times in metrics registration, but we don't have strong preference on whether we should make `variants` an instance attribute or just an input parameter of `generate_p_value`. I went with the latter one, because to me `variants` is similar to experiment-level parameters like alpha and power, and there's no need to make them as each metric class's instance attribute

sample_size/metrics.py
xyzou685: This is something we can discuss. They are using z test as shown in line 77, but I still use `nct` when sampling their test statistics, because Kolmogorov–Smirnov tests' p-values are not uniformly distributed at all when I replaced

```
def direct_sample(δ, σ, n, size):
    nc = np.sqrt(n/2) * δ / σ
    return nct.rvs(nc=nc, df=2*(n-1), size=size)
```
by
```
def direct_sample(δ, σ, n, size):
    effect_size = δ / float(np.sqrt(2*σ**2/n))
    return stats.norm.rvs(loc=effect_size, size=size)
```
 in your script

tests/sample_size/test_metrics.py
xyzou685: I got `AttributeError: can't set attribute` when doing this

sample_size/metrics.py
```diff
@@ -60,6 +67,15 @@ def _check_probability(probability: float) -> float:
         else:
             raise ValueError("Error: Please provide a float between 0 and 1 for probability.")
 
+    def generate_p_value(self, true_alt: bool, size: int, variants: int) -> Any:
```
pchiang-godaddy, line 70: can we avoid Any as the return type? same for other ones

sample_size/metrics.py
pchiang-godaddy, line 38: it is not clear to me to directly know the meaning of each parameters for `true_alt`, `size`, and `variants`.  Maybe worth a docstring?
alt = alternative?

dfu-godaddy: +1. or we can make the variables names longer and clearer

sample_size/metrics.py
xyzou685: Originally I used `Union[float, List[float]]` as all `generate_p_value` return type, but I always got the following type check error:
`Returning Any from function declared to return "Union[float, List[float]]"  [no-any-return]`

the actual return type of this function is either one p-value(when variants=2) or a list of p-values

pchiang-godaddy: the problem is probably because np function didn't indicate a type so mypy doesn't know its type. you can explicitly provide it e.g.

```python
    def generate_p_value(self, true_alt: bool, size: int, variants: int) -> Union[np.typing.NDArray[np.float64], float]:
        if not true_alt:
            return np.array(stats.uniform.rvs(0, 1, size=variants - 1))

        else:
            nc = np.sqrt(size / 2) * self.mde / self.variance
            z_alt = stats.nct.rvs(nc=nc, df=2 * (size - 1), size=variants - 1)
            return float(2 * stats.norm.sf(np.abs(z_alt)))
```

and update your test to be 
```python
from numpy.testing import assert_array_equal
assert_array_equal(p, pval)
```

tests/sample_size/test_metrics.py
pchiang-godaddy, line 146:
```suggestion
        assert_array_equal(p, pval)
```
this is more readable and can pass mypy https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

xyzou685: I got the same error even after replacing the type by  `Union[np.typing.NDArray[np.float64], float]`. The only way to have the tests pass is making the returned variable to be an ndarray:
```
  def generate_p_value(self, true_null: bool, sample_size: int, variants: int) -> np.typing.NDArray[np.float64]:
      if true_null:
          return np.array(stats.uniform.rvs(0, 1, size=variants - 1))

      else:
          effect_size = self.mde / float(np.sqrt(2 * self.variance / sample_size))
          z_alt = stats.norm.rvs(loc=effect_size, size=variants - 1)
          return np.array(2 * stats.norm.sf(np.abs(z_alt)))
```
btw I also made changes to the variable names and added a docstring

pchiang-godaddy: yeah my suggestion includes making the type explicitly in this comment https://github.com/godaddy/sample-size/pull/26#discussion_r875242493, btw can you use `assert_array_equal(p, pval)` instead of `(p == pval).all()`?
with your above code, idk if you need to change anything else if both conditions return np.array. `Union[np.typing.NDArray[np.float64], float]` this means it returns numpy array or a float. But in general i guess it makes more sense to return the same type probably (?) curious to know what others think

BTW it seems like you didn't commit the change yet

xyzou685: There's still issue with typing. I just committed the change along with other unit test changes. It failed the checks since I kept getting the error `AttributeError: module 'numpy' has no attribute 'typing'`

vjohnson1-godaddy: Numpy has some some [a special type plugin](https://numpy.org/devdocs/reference/typing.html). Perhaps that's the issue?

pchiang-godaddy: you can check the error under Actions tab https://github.com/godaddy/sample-size/runs/6498722018?check_suite_focus=true
it is the package problem, i think you need to also include lock file?

xyzou685:
Since we don't necessarily have to get `np.ndarray` as the output type, I circumvented the use of it by make them `List[float]`. 

It that seems fine, the last question I want to discuss is the value check in unit test: currently I am doing 
```
        if true_null:
            mock_uniform.assert_called_once_with(0, 1, size=variants - 1)
            mock_norm.assert_not_called()
            assert_array_equal(p, pval)
        
        else:
            effect_sample_size = self.DEFAULT_MDE / float(np.sqrt(2 * self.DEFAULT_VARIANCE / sample_size))
            mock_uniform.assert_not_called()
            mock_norm.assert_called_once_with(loc=effect_sample_size, size=variants - 1)
            self.assertEqual(len(p), len(pval))
```
instead of 
```
        if true_null:
            mock_uniform.assert_called_once_with(0, 1, size=variants - 1)
            mock_norm.assert_not_called()
            
        else:
            effect_sample_size = self.DEFAULT_MDE / float(np.sqrt(2 * self.DEFAULT_VARIANCE / sample_size))
            mock_uniform.assert_not_called()
            mock_norm.assert_called_once_with(loc=effect_sample_size, size=variants - 1)
        assert_array_equal(p, pval)
```
because `mock_norm` is mocking `norm.rvs`(whose returned value is expected to be `pval`), not `norm.sf`, so the function's output in this case is not going to match the preset `pval`. Therefore in the case of true alternative, I only checked if the lengths of the output match. Is this sufficient for testing `generate_p_value`?

pyproject.toml
```diff
@@ -17,10 +17,12 @@ flake8 = "^4.0"
 coverage = "^6.3"
 isort = "^5.6.3"
 black = "^21.7b0"
-mypy = "^0.942"
+mypy = "^0.950"
 pytest = "^7.0.0"
 pytest-sugar = "^0.9.4"
 pytest-cov = "^3.0.0"
+click = "8.0.4"
```
pchiang-godaddy, line 24: nit: probably update branch of [multiple-testing-calculator](https://github.com/godaddy/sample-size/tree/multiple-testing-calculator), so we know it is not an update you made in this PR?

tests/sample_size/test_metrics.py
```diff
@@ -76,6 +80,29 @@ def test_boolean_metric_get_probability_too_small(self):
...
+
+        boolean = BooleanMetric(self.DEFAULT_PROBABILITY, self.DEFAULT_MDE)
+        p = boolean.generate_p_value(true_null, sample_size, variants)
+
+        if true_null:
+            mock_uniform.assert_called_once_with(0, 1, size=variants - 1)
+            mock_norm.assert_not_called()
+            assert_array_equal(p, pval)
+        else:
+            effect_sample_size = self.DEFAULT_MDE / float(np.sqrt(2 * self.DEFAULT_MOCK_VARIANCE / sample_size))
```
vjohnson1-godaddy, line 101: nit: shouldn't need the float here
```suggestion
            effect_sample_size = self.DEFAULT_MDE / np.sqrt(2 * self.DEFAULT_MOCK_VARIANCE / sample_size)
```

tests/sample_size/test_metrics.py
```diff
@@ -76,6 +80,29 @@ def test_boolean_metric_get_probability_too_small(self):
...
+            mock_uniform.assert_not_called()
+            mock_norm.assert_called_once_with(loc=effect_sample_size, size=variants - 1)
+            self.assertEqual(len(p), len(pval))
```
vjohnson1-godaddy, line 104: I think with some diffrent mocking you can test the value of what the p-value generator returns in both cases
```suggestion
    @parameterized.expand(product([True, False], [2, 5], [200]))
    @patch("sample_size.metrics.BooleanMetric.variance")
    @patch("scipy.stats.uniform.rvs")
    @patch("scipy.stats.norm")
    def test_boolean_generate_p_value(
        self, true_null, variants, sample_size, mock_norm, mock_uniform_rvs, mock_variance
    ):
        mock_norm.rvs.return_value = -ord("🌮")
        p_value_generator = mock_uniform_rvs if true_null else mock_norm.sf
        pval = np.zeros(variants - 1)
        p_value_generator.return_value = pval
        mock_variance.__get__ = MagicMock(return_value=self.DEFAULT_MOCK_VARIANCE)

        boolean = BooleanMetric(self.DEFAULT_PROBABILITY, self.DEFAULT_MDE)
        p = boolean.generate_p_value(true_null, sample_size, variants)

        if true_null:
            mock_uniform_rvs.assert_called_once_with(0, 1, size=variants - 1)
            mock_norm.assert_not_called()
        else:
            effect_sample_size = self.DEFAULT_MDE / float(np.sqrt(2 * self.DEFAULT_MOCK_VARIANCE / sample_size))
            mock_uniform_rvs.assert_not_called()
            mock_norm.rvs.assert_called_once_with(loc=effect_sample_size, size=variants - 1)
            mock_norm.sf.assert_called_once_with(np.abs(mock_norm.rvs.return_value))

        assert_array_equal(p, pval)
```

sample_size/metrics.py
```diff
@@ -31,6 +34,21 @@ def check_positive(number: float, name: str) -> float:
...
+        """
+        This method simulates registered metric's p-value(s). The output will later be applied to BH procedure
+
+        Parameters:
+            true_null: whether the null hypothesis is true
+            sample_size: sample size used for simulations
+            variants: number of test variants, including control
+
+        Returns:
+            p-value(s): the p-values generated. This is a float or a list with length variants-1
```
dfu-godaddy, line 48: nit: The return is set to a list. Would it ever be a float?

xyzou685: It can be a list with one element but is always a list. I have changed this docstring to avoid confusion

.github/workflows/publish.yml
```diff
@@ -8,9 +8,9 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+    - uses: actions/checkout@d50f8ea76748df49594d9b109b614f3b4db63c71
     - name: Set up Python 3.7
-      uses: actions/setup-python@c57f79353baa47cac20975f0a134915e0a8c18bf
+      uses: actions/setup-python@6c048c755810cf9ec19b2a2a75014fcd1aa0c9c6
```
pchiang-godaddy, line 13: can you remove your change here?

pchiang-godaddy (changes requested):

pchiang-godaddy (commented):
you can merge after `.github/workflows/publish.yml` file change is reverted

PR #30, opened 2022-05-20
feat(multiple_testing_power): make input_utils support multiple metrics

xyzou685:
To help us get this pull request reviewed and merged quickly, please be sure to include the following items:

* [x] Tests (if applicable)
* [ ] Documentation (if applicable)
* [ ] A full explanation here in the PR description of the work done and statistical methods used

## PR Type
What kind of change does this PR introduce?

* [ ] Bugfix
* [x] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation content changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

Is this change backward compatible with the most recently released version? Does it introduce changes which might change the user experience in any way? Does it alter the API in any way?

* [ ] Yes (backward compatible)
* [ ] No (breaking changes)

PR #32, opened 2022-05-27
feat(multiple_testing_power): add multiple testing functions

xyzou685:

* [x] Tests (if applicable)

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [x] No (breaking changes)


dfu-godaddy (commented):
can you rebase your branch after you merge https://github.com/godaddy/sample-size/pull/30 so the earlier updates are not showing in this PR

sample_size/sample_size_calculator.py
```diff
@@ -22,43 +110,19 @@ class SampleSizeCalculator:
...
-        covariance: float,
-        mde: float,
-    ) -> None:
-        metric = RatioMetric(
-            numerator_mean, numerator_variance, denominator_mean, denominator_variance, covariance, mde
-        )
-        self.ratio_metrics.append(metric)
-
-    def _get_single_sample_size(self, metric: BaseMetric) -> float:
+    def get_single_sample_size(self, metric: BaseMetric, alpha=self.alpha) -> float:
```
dfu-godaddy, line 119: is `alpha=self.alpha` a good practice? my pycharm gives me a compiling error

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    metrics: List[BaseMetric]
+    alpha: float
+    power: float
+    variants: int
+
+    def get_multiple_sample_size(self) -> int:
+        if len(self.metrics) < 2:
+            return self.get_single_sample_size(self.metrics[0])
+        lower = max([self.get_single_sample_size(metric) for metric in self.metrics])
```
dfu-godaddy, line 27: `self.get_single_sample_size` does not exist in `MultipleTestingMixin`, I don't think we should call it here

xyzou685: > can you rebase your branch after you merge #30 so the earlier updates are not showing in this PR

Actually I did pull most changes from #30, otherwise it's blocking this PR. I will sync the last few changes after its merged

pchiang-godaddy:
+1 on updating the branch to the latest base branch, and the error is from lint in the check now
```
./tests/sample_size/scripts/test_multiple_testing.py:5:1: F401 'numpy.testing.assert_equal' imported but unused
./tests/sample_size/scripts/test_multiple_testing.py:7:1: F401 'sample_size.metrics.BooleanMetric' imported but unused
./tests/sample_size/scripts/test_multiple_testing.py:8:1: F401 'sample_size.metrics.NumericMetric' imported but unused
./tests/sample_size/scripts/test_multiple_testing.py:9:1: F401 'sample_size.metrics.RatioMetric' imported but unused
./tests/sample_size/scripts/test_multiple_testing.py:[10](https://github.com/godaddy/sample-size/runs/6675979787?check_suite_focus=true#step:8:11):1: F401 'sample_size.multiple_testing.MultipleTestingMixin' imported but unused
./tests/sample_size/scripts/test_multiple_testing.py:[11](https://github.com/godaddy/sample-size/runs/6675979787?check_suite_focus=true#step:8:12):1: F401 'sample_size.sample_size_calculator.DEFAULT_ALPHA' imported but unused
./tests/sample_size/scripts/test_multiple_testing.py:[12](https://github.com/godaddy/sample-size/runs/6675979787?check_suite_focus=true#step:8:13):1: F401 'sample_size.sample_size_calculator.DEFAULT_POWER' imported but unused
```

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if np.isclose((self.power, expected_power), EPSILON) or np.isclose((candidate, upper), EPSILON):
+            return candidate
+
+        if expected_power > self.power:
+            return self.get_multiple_sample_size(lower, candidate, depth + 1)
+        else:
+            return self.get_multiple_sample_size(candidate, upper, depth + 1)
+
+    def _expected_average_power(self, sample_size: int, REPLICATION: int = 100) -> float:
```
dsun1-godaddy, line 44: nit: don't think a function parameter should be all caps. Pull REPLICATION into class variable and change the parameter to replication: int = REPLICATION

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    metrics: List[BaseMetric]
+    alpha: float
+    power: float
+    variants: int
+
+    def get_multiple_sample_size(self, lower: float, upper: float, depth: int = 0) -> int:
+        MAX_RECURSION_DEPTH = 20
+        EPSILON = 0.025
```
vjohnson1-godaddy, line 28: Could you make a JIRA ticket for this?

```suggestion
        EPSILON = 0.025 #  TODO(any): make this configurable by the user ML-XXXX
```

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if depth > MAX_RECURSION_DEPTH:
+            raise RecursionError
```
vjohnson1-godaddy, line 31:
```suggestion
            raise RecursionError(f"Couldn't find a sample size that satisfies the power you requested: {self.power}")
```

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,57 @@
...
+
+        candidate = int((upper + lower) / 2)
+        expected_power = self._expected_average_power(candidate)
+
+        if np.isclose((self.power, expected_power), EPSILON) or np.isclose((candidate, upper), EPSILON):
+            return candidate
```
vjohnson1-godaddy, line 37: By definition, if we're evaluating the latter condition, we know `candidate` is not close to the true sample size, so we shouldn't return it. Technically, you don't need to perform this calculation at all (the recursion depth will implicitly handle this case), but if you would like to end the recursion early, you can add this second condition.

```suggestion
        if np.isclose((self.power, expected_power), EPSILON):
            return candidate
       
        if upper == lower:
            raise RuntimeError("f"Couldn't find a sample size that satisfies the power you requested: {self.power}"
```

tests/sample_size/scripts/test_multiple_testing.py
```diff
@@ -0,0 +1,104 @@
...
+        test_metric_type = "boolean"
+        test_probability = 0.05
+        test_mde = 0.02
+        test_metric_metadata = {"probability": test_probability, "mde": test_mde}
+        calculator = SampleSizeCalculator()
+        calculator.register_metrics(
+            [
+                {"metric_type": test_metric_type, "metric_metadata": test_metric_metadata},
+                {"metric_type": test_metric_type, "metric_metadata": test_metric_metadata},
+            ]
```
vjohnson1-godaddy, line 23: Might be good move almost all of this code to a [`setUp` function](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) that can be reused across your tests

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,57 @@
...
+    variants: int
+
+    def get_multiple_sample_size(self, lower: float, upper: float, depth: int = 0) -> int:
+        MAX_RECURSION_DEPTH = 20
+        EPSILON = 0.025
+
+        if depth > MAX_RECURSION_DEPTH:
+            raise RecursionError
+
+        candidate = int((upper + lower) / 2)
```
pchiang-godaddy, line 33:
```suggestion
        candidate = int(upper + (upper-lower) / 2)
```
is upper and lower "size"? maybe try to write it in this way to avoid integer out of bound if it could be a very large number

sample_size/multiple_testing.py
vjohnson1-godaddy: I think that concern would be best addressed by whatever client is calling this function. It should probably be sufficient to just raise the exception

dsun1-godaddy (approved):

tests/sample_size/test_sample_size_calculator.py
```diff
@@ -81,83 +82,82 @@ def test_get_single_sample_size_ttest(self, mock_solve_power):
...
+    ):
+        test_metric_type = metric_type
         test_sample_size = 2000
         mock_get_single_sample_size.return_value = test_sample_size
 
-        test_mde = 0.02
-        test_probability = 0.05
-        test_metric_metadata = {"probability": test_probability, "mde": test_mde}
+        # test_mde = 0.02
+        # test_probability = 0.05
```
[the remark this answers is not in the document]
xyzou685: Do you know how I can loop over each metric class's properties? for example, `assert_equal(mock_get_single_sample_size.call_args[0][0].probability, test_probability)` is only for boolean and is not applicable to other metric classes

sample_size/multiple_testing.py
pchiang-godaddy: guess we never know if it will be an issue or not? just provide a way so we don't need to worry about that

tests/sample_size/test_sample_size_calculator.py
```diff
@@ -81,83 +82,82 @@ def test_get_single_sample_size_ttest(self, mock_solve_power):
...
-        test_metric_metadata = {
-            "numerator_mean": test_numerator_mean,
-            "numerator_variance": test_numerator_variance,
-            "denominator_mean": test_denominator_mean,
-            "denominator_variance": test_denominator_variance,
-            "covariance": test_covariance,
-            "mde": test_mde,
-        }
-
+        # test_mde = 0.02
```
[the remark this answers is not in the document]
xyzou685: I'm trying to add these back https://github.com/godaddy/sample-size/pull/32#discussion_r889166314. Or are we good to actually skip the metadata check for now? cc @pchiang-godaddy @vjohnson1-godaddy

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,59 @@
...
+    metrics: List[BaseMetric]
+    alpha: float
+    power: float
+    variants: int
+
+    REPLICATION: int = 100
+
+    def get_multiple_sample_size(self, lower: float, upper: float, depth: int = 0) -> int:
+        max_recursion_depth: int = 20
+        epsilon: float = 0.025  # TODO(any): make this configurable by users ML-5429
```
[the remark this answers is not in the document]
xyzou685: I didn't put them as default attributes since they are used only in this method. 

Yes, I just created it: https://github.com/godaddy/sample-size/pull/35

tests/sample_size/test_sample_size_calculator.py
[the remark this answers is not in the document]
vjohnson1-godaddy: I think you can get rid of this it now because it's capture in the metadata that you parameterized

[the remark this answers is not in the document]
pchiang-godaddy: since you actually validate the metric by `calculator.metrics[0]` i'm fine skipping them. but can you add one more check for call count? since `has_calls` doesn't require you to provide all the calls
```python
self.assertEqual(mock_get_single_sample_size.call_count, 4)
```

tests/sample_size/test_sample_size_calculator.py
```diff
@@ -81,83 +82,82 @@ def test_get_single_sample_size_ttest(self, mock_solve_power):
...
         sample_size = calculator.get_sample_size()
 
         self.assertEqual(sample_size, test_sample_size)
         mock_get_single_sample_size.assert_called_once()
-        self.assertIsInstance(mock_get_single_sample_size.call_args[0][0], BooleanMetric)
-        assert_equal(mock_get_single_sample_size.call_args[0][0].probability, test_probability)
-        assert_equal(mock_get_single_sample_size.call_args[0][0].mde, test_mde)
+        mock_get_multiple_sample_size.assert_not_called()
+        self.assertIsInstance(mock_get_single_sample_size.call_args[0][0], metric_class)
+        # assert_equal(mock_get_single_sample_size.call_args[0][0].probability, test_probability)
```
pchiang-godaddy, line 124: I'm fine if we do below instead without expllicitly validating each parameters under each metric 
```python
self.assertEqual(mock_get_single_sample_size.call_args[0][0], calculator.metrics[0])
```

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,59 @@
...
+            return candidate
+
+        if expected_power > self.power:
+            return self.get_multiple_sample_size(lower, candidate, depth + 1)
+        else:
+            return self.get_multiple_sample_size(candidate, upper, depth + 1)
+
+    def _expected_average_power(self, sample_size: int, replication: int = REPLICATION) -> float:
+        power = []
+        m = len(self.metrics) * (self.variants - 1)
```
dfu-godaddy, line 48: can we rename `m` to something more user friendly, like `num_of_tests` or `p_value_count`?

sample_size/sample_size_calculator.py
```diff
@@ -49,7 +50,11 @@ def _get_single_sample_size(self, metric: BaseMetric) -> float:
         return sample_size
 
     def get_sample_size(self) -> float:
-        return self._get_single_sample_size(self.metrics[0])
+        if len(self.metrics) < 2:
```
dfu-godaddy, line 53: we are not updating the sample size if the user has 1 metric but 5 variants?

dfu-godaddy (commented):
Tried the script a couple of times for multiple metrics on my local. all failed. Can you check this?
image
image

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,67 @@
...
+                                sample_size,
+                            )
+                            for i, m in enumerate(self.metrics)
+                        ]
+                    )
+
+                rejected = multipletests(p_values, alpha=self.alpha, method="fdr_bh")[0]
+                power.append(sum(rejected[~true_null]) / (num_of_tests - m1))
+
+        return float(np.mean(power))
```
vjohnson1-godaddy, line 67: Might be a little more legible this way
```suggestion
        for num_true_null in range(num_of_tests):
            num_true_alt = num_of_tests - num_true_null
            nulls = np.array([True] * num_true_null + [False] * (num_true_alt))
            for _ in range(replication):
                p_values = []
                true_null = nulls[np.random.permutation(num_of_tests)]
                for v in range(self.variants - 1):
                    p_values.extend(
                        [
                            m.generate_p_value(
                                true_null[v * len(self.metrics) + i],
                                sample_size,
                            )
                            for i, m in enumerate(self.metrics)
                        ]
                    )

                rejected = multipletests(p_values, alpha=self.alpha, method="fdr_bh")[0]
                power.append(sum(rejected[~true_null]) / (num_true_alt))

        return float(np.mean(power))
```

tests/sample_size/scripts/test_multiple_testing.py
```diff
@@ -0,0 +1,97 @@
...
+        "sample_size.sample_size_calculator.SampleSizeCalculator._get_single_sample_size",
+        side_effect=[100, 100, 1000, 1000],
+    )
+    @patch("sample_size.multiple_testing.MultipleTestingMixin._expected_average_power", side_effect=[0.5, 0.8])
+    def test_get_multiple_sample_size_too_low(self, mock_expected_average_power, mock_get_single_sample_size):
+        calculator = SampleSizeCalculator()
+        calculator.register_metrics([self.test_metric, self.test_metric])
+        sample_size = calculator.get_sample_size()
+        self.assertEqual(mock_get_single_sample_size.call_count, 4)
+        self.assertEqual(mock_expected_average_power.call_count, 2)
```
vjohnson1-godaddy, line 27: Could help clarify why we're asserting 2 and 4 here? It's not clear to me why these are _supposed_ to be 2 and 4.

tests/sample_size/scripts/test_multiple_testing.py
```diff
@@ -0,0 +1,97 @@
...
+        calculator = SampleSizeCalculator()
+        calculator.register_metrics([self.test_metric, self.test_metric])
+
+        with self.assertRaises(Exception) as context:
+            calculator.get_sample_size()
+
+        self.assertEqual(
+            str(context.exception),
+            f"Couldn't find a sample size that satisfies the power you requested: {test_power}",
+        )
```
vjohnson1-godaddy, line 78: To the extend that we're using `RuntimeError` rather than rolling our own `Exception` class, I don't think we need this. If we changed the error message, I would want the unit tests to still pass
```suggestion
```

tests/sample_size/scripts/test_multiple_testing.py
```diff
@@ -0,0 +1,97 @@
...
+        self, mock_expected_average_power, mock_get_single_sample_size
+    ):
+        test_power = 0.8
+        mock_expected_average_power.return_value = 0.1
+
+        calculator = SampleSizeCalculator()
+        calculator.register_metrics([self.test_metric, self.test_metric])
+
+        with self.assertRaises(Exception) as context:
+            calculator.get_sample_size()
```
vjohnson1-godaddy, line 73:
```suggestion
    @patch("sample_size.multiple_testing.MultipleTestingMixin.np.is_close", False)
    def test_get_multiple_sample_size_no_solution(self):
        calculator = SampleSizeCalculator()
        calculator.register_metrics([self.test_metric, self.test_metric])

        with self.assertRaises(Exception) as context:
            calculator.get_sample_size()
```

tests/sample_size/scripts/test_multiple_testing.py
```diff
@@ -0,0 +1,97 @@
...
+        side_effect=[100, 100, 1000, 1000],
+    )
+    @patch("sample_size.multiple_testing.MultipleTestingMixin._expected_average_power", side_effect=[1, 0.8])
+    def test_get_multiple_sample_size_too_high(self, mock_expected_average_power, mock_get_single_sample_size):
+        calculator = SampleSizeCalculator()
+        calculator.register_metrics([self.test_metric, self.test_metric])
+        sample_size = calculator.get_sample_size()
+        self.assertEqual(mock_get_single_sample_size.call_count, 4)
+        self.assertEqual(mock_expected_average_power.call_count, 2)
+        self.assertEqual(sample_size, ((self.test_upper + self.test_lower) / 2 + self.test_lower) / 2)
```
vjohnson1-godaddy, line 41:
```suggestion
    @paramterized.expand([0, 1])
    @patch(
        "sample_size.sample_size_calculator.SampleSizeCalculator._get_single_sample_size",
        side_effect=[100, 100, 1000, 1000],
    )
    @patch("sample_size.multiple_testing.MultipleTestingMixin._expected_average_power")
    def test_get_multiple_sample_size_quickly_converge(self, mock_expected_average_power, mock_get_single_sample_size, power_guess):
        mock_expected_average_power.side_effect = [power_guess, 0.8]
        calculator = SampleSizeCalculator()
        calculator.register_metrics([self.test_metric, self.test_metric])
        sample_size = calculator.get_sample_size()
        self.assertEqual(mock_get_single_sample_size.call_count, 4)
        self.assertEqual(mock_expected_average_power.call_count, 2)
        self.assertEqual(sample_size, ((self.test_upper + self.test_lower) / 2 + self.test_lower) / 2)
```

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,67 @@
...
+        max_recursion_depth: int = 20
+        epsilon: float = 0.025  # TODO(any): make this configurable by users ML-5429
+
+        if depth > max_recursion_depth:
+            raise RecursionError(f"Couldn't find a sample size that satisfies the power you requested: {self.power}")
+
+        candidate = int(lower + (upper - lower) / 2)
+        expected_power = self._expected_average_power(candidate)
+        if np.isclose(self.power, expected_power, epsilon) or (lower == upper and self.power < expected_power):
+            return candidate
```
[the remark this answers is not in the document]
xyzou685: I added the condition `lower == upper and self.power < expected_power` because in some cases like https://github.com/godaddy/sample-size/pull/32#pullrequestreview-995494156, the lower bound sample size generates average power way above 0.8. In this case we do want to use lower bound, i.e. the maximum of individual sample sizes, instead of raising a runtime error

tests/sample_size/scripts/test_multiple_testing.py
xyzou685: Every time we run multiple testing calculator, we are calculating each metric's required sample size and get the lower and upper bound. Thus, `mock_get_single_sample_size` will be called for `len(metrics)*2` times. In this test case we have 2 recursions so `mock_expected_average_power` is expected to be called twice.

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if expected_power > self.power:
+            return self.get_multiple_sample_size(lower, candidate, depth + 1)
+        else:
+            return self.get_multiple_sample_size(candidate, upper, depth + 1)
+
+    def _expected_average_power(self, sample_size: int, replication: int = REPLICATION) -> float:
+        power = []
+        num_of_tests = len(self.metrics) * (self.variants - 1)
+        for m1 in range(num_of_tests):
```
pchiang-godaddy, line 48: do we have any paper formula on the below calculation logic? might be good to link it here(if we have) so it is easier to know what it is doing.

sample_size/multiple_testing.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    metrics: List[BaseMetric]
+    alpha: float
+    power: float
+    variants: int
+
+    REPLICATION: int = 500
+
+    def get_multiple_sample_size(self, lower: float, upper: float, depth: int = 0) -> int:
```
pchiang-godaddy, line 28: nit - probably can add a docstring for readability - what each parameter (lower/upper) means for example. not easy to know lower and upper represent "sample size" when reading directly

sample_size/multiple_testing.py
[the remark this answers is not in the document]
V-Neck: Oh! I see the bug! I think we proved in the simulation studies that BH uniformly decreases power (by definition it can't lower it because it is conservative, and never decreases p-values), so we shouldn't never run into the problem you're describing, but I think

```python
        lower = max([self._get_single_sample_size(metric, self.alpha) for metric in self.metrics])
```

should be

```python3
        lower = min([self._get_single_sample_size(metric, self.alpha) for metric in self.metrics])
```

I suspect that would fix the lower bound still being > 0.8 problem

tests/sample_size/scripts/test_multiple_testing.py
V-Neck: I see. This dovetails with my comment below. Perhaps we could parameterize the number of recursive calls?

```python
    @paramterized.expand(product([0, 1], [1, 2, 3]))
    @patch(
        "sample_size.sample_size_calculator.SampleSizeCalculator._get_single_sample_size",
        side_effect=[100, 100, 1000, 1000],
    )
    @patch("sample_size.multiple_testing.MultipleTestingMixin._expected_average_power")
    def test_get_multiple_sample_size_quickly_converge(self, mock_expected_average_power, mock_get_single_sample_size, power_guess, n_recursive_calls):
        mock_expected_average_power.side_effect = ([power_guess] * n_recursive_calls) + [0.8]
        calculator = SampleSizeCalculator()
        calculator.register_metrics([self.test_metric, self.test_metric])
        sample_size = calculator.get_sample_size()
        self.assertEqual(mock_get_single_sample_size.call_count, 2 * n_recursive_calls)
        self.assertEqual(mock_expected_average_power.call_count, n_recursive_calls)
        self.assertEqual(sample_size, ((self.test_upper + self.test_lower) / 2 + self.test_lower) / 2)
```

vjohnson1-godaddy: ^whoops, logged into wrong github account

tests/sample_size/scripts/test_multiple_testing.py
```diff
@@ -0,0 +1,97 @@
...
+        mock_multipletests.return_value = [1, 1, 1]
+        mock_isclose.return_value = True
+
+        calculator = SampleSizeCalculator()
+        calculator.register_metrics([self.test_metric, self.test_metric, self.test_metric])
+        replication = len(calculator.metrics) ** 2 * (calculator.variants - 1) * 500
+
+        sample_size = calculator.get_multiple_sample_size(self.test_lower, self.test_upper)
+        self.assertEqual(mock_generate_p_value.call_count, replication)
+        self.assertEqual(sample_size, ((self.test_upper + self.test_lower) / 2))
```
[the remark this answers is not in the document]
xyzou685: I changed it to solely run `_expected_average_power` and check that for each sample size, whether inflating by 10 times will generate higher power. I also removed mocking `generate_p_value` since we do need its value to calculate average power and thus check if the power >0.8

sample_size/multiple_testing.py
xyzou685: We don't have exact paper formula on this, but it's derived from the formula in this [PR](https://github.secureserver.net/hivemind/pdrs/blob/master/hivemind-power-analysis/Model_the_effect_BH_correction_has_on_Power.md#deriving-average-power-and-sample-size-algebraically). For now I added a docstring on each function in `multiple_testing`

tests/sample_size/scripts/test_multiple_testing.py
xyzou685: deleting this assertion will fail to make unit test have 100% coverage

vjohnson1-godaddy: Okie! Sounds good!

sample_size/multiple_testing.py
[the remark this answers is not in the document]
xyzou685: It did fix the problem!

vjohnson1-godaddy (approved):

PR #35, opened 2022-06-03
feat(multiple_testing_power): support multiple metrics and cohort sample size calculation

xyzou685:
To help us get this pull request reviewed and merged quickly, please be sure to include the following items:

* [x] Tests (if applicable)
* [x] Documentation (if applicable)
* [ ] A full explanation here in the PR description of the work done and statistical methods used

## PR Type
What kind of change does this PR introduce?

* [ ] Bugfix
* [x] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation content changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

Is this change backward compatible with the most recently released version? Does it introduce changes which might change the user experience in any way? Does it alter the API in any way?

* [ ] Yes (backward compatible)
* [x] No (breaking changes)

PR #41, opened 2022-08-02
perf: General optimizations and parameter tuning research

vjohnson1-godaddy:
# Background
Following PR #35, there has been room for optimization. This PR contains the research needed in order to subsequently introduce some of those optimizations

# Change List
- A new notebook
- code used in notebook to compare performance before and after optimization


vjohnson1-godaddy:
I'm not sure what the best course of action for the SonarCloud check is. It's yelling because I've duplicated parts of the current version into the `notebooks/` directory in order to make a before and after comparison of performance.

Perhaps I should instead include the diff? I can continue to use the CSVs to populate the results of the notebook (will complicate reproducibility, though

dfu-godaddy (commented):
Could we have the calculator, metrics, and multiple_testing updated in the same file so we can have a better idea of what is changed in this PR? Besides, can we keep the performance comparison as an internal PDR?

vjohnson1-godaddy:
> Besides, can we keep the performance comparison as an internal PDR?

I think the parameter tuning should be included here, but I can live without including the other stuff. The reason the rest of the optimizations were included was because it would have been impractical to tune those parameters without faster code. I'll just close the PR and we can return to the parameter after it's merged

vjohnson1-godaddy:
I'll just put out a regular PR for the numpy optimizations, and we can circle back to the notebook after that's merged to justify the choice of parameters

